PR #20, opened 2017-11-13
RLM-355 Add Vagrantfile for Testing rpc-upgrades

antonym:
Adds an initial vagrantfile


claco:
Merging. The old jobs are pre-renaming. The two current PR jobs passed.

PR #22, opened 2017-11-14
RE-913 Add support for minor upgrades

git-harry:
The current minor upgrade path simply involves running the newer version
of the deployment script, there is no specific upgrade code path to run.

This change introduces `tests/test-minor.sh` for running a minor upgrade
and `tests/test-upgrade.sh` for controlling which type of upgrade is
performed. If RE_JOB_UPGRADE_ACTION is either "leap" or "minor" the
appropriate script is executed.

`tests/aio-create.sh` is updated to support deploying a newton version
of the code as the base so that minor upgrades on the current release
can be tested.


git-harry:
recheck_cit_all

PR #25, opened 2017-11-15
RLM-294 Integrate preflight check into gating

Frank-ZhangXin:
Adding prelight check prior any functional check in gating.


Frank-ZhangXin:
Closed because duplicated using in https://github.com/rcbops/rpc-upgrades/pull/24

PR #27, opened 2017-11-16
RLM-382 Checkout branch directly instead of master

antonym:
Checks out the target branch directly to avoid any sort of submodule collisions by checking out master first.


antonym:
PR jobs got past broken point so merging.

PR #28, opened 2017-11-16
RLM-305 Enable MaaS Deploy and Verify before Leap

antonym:
Deploys and verifies MaaS after deploy of AIO and before leapfrog.  

Uses playbooks in the repo copied from head rpc-openstack to checkout rpc-maas repo.  The older point releases of the various versions have the old maas release built in and is not using the rpc-maas repo.  Since the rpc-maas repo is our current preferred method, this runs and installs maas from there.

### DIFF
--- a/playbooks/ansible.cfg
+++ b/playbooks/ansible.cfg
@@ -1,0 +1,45 @@
+[defaults]
+gathering = smart
+host_key_checking = False
+
+# Setting forks should be based on your system. The Ansible defaults to 5,
+# the os-lxc-hosts assumes that you have a system that can support
+# OpenStack, thus it has been conservatively been set to 15
+forks = 15
+
+# SSH timeout
+timeout = 120
+
+# Set the path to the folder in openstack-ansible which holds the dynamic
+# inventory script - new config setting for ansible v1.9 and above
+inventory = /opt/rpc-openstack/openstack-ansible/playbooks/inventory/
+
+# Set the path to the folder in openstack-ansible which holds the dynamic
+# inventory script - uncomment if using ansible below v1.9
+#hostfile = ../../openstack-ansible/playbooks/inventory/
+
+# Set the path to the folder in openstack-ansible which holds the
+# libraries required
+library = /opt/rpc-openstack/openstack-ansible/playbooks/library/
+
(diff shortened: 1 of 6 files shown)


antonym:

playbooks/ansible.cfg
```diff
@@ -0,0 +1,45 @@
+[defaults]
```
cloudnull: I don't think we need this file. If we run the playbooks with the `openstack-ansible` command line wrapper everything that we should need should be provided, from a library, module, role perspective.

playbooks/maas-verify.yml
```diff
@@ -0,0 +1,16 @@
+---
```
cloudnull: once we have maas in the /opt/rpc-maas location there's no need to wrap this the verify playbook here.

tests/maas-install.sh
```diff
@@ -0,0 +1,26 @@
...
+
+## Shell Opts ----------------------------------------------------------------
+
+set -evu
+
+pushd /opt/rpc-upgrades/playbooks
+  # install maas
+  openstack-ansible maas-setup.yml -vv
+  # verify maax is running
+  openstack-ansible maas-verify.yml -vv
```
cloudnull: This should just be `/opt/rpc-maas/playbooks/maas-verify.yml` there's no need to warp the existing playbook.

playbooks/ansible.cfg
antonym: Was having issues running playbooks from within the rpc-upgrades repo (couldn't find inventory file) which is why I added it.  Once I added it, it seemed to link up everything correctly.

tests/maas-install.sh
antonym: Yeah, makes sense, will fix up.

playbooks/maas-verify.yml
antonym: Cool, works for me.

playbooks/ansible.cfg
cloudnull: OK

antonym:
@cloudnull yeah, the failing lint is unrelated to this commit.  Opened a card to circle back and get those fixed: https://rpc-openstack.atlassian.net/browse/RLM-416

PR #32, opened 2017-11-20
RLM-301 Migrate all stuff of ES upgrade from RPC-O

Frank-ZhangXin:
Improve scripts to adapt RPC-upgrades and automate process of
Elasticsearch upgrade(preserving ES logs).


scripts/pre_leap.sh
```diff
@@ -19,6 +19,16 @@
 set -e -u
 set -o pipefail
 
+# Choose if preserve Elasticsearch data in the leap
+read -p 'Enter "YES" to make sure you want to preserve Elasticsearch logs in leapfrog or anything else to skip: ' TO_UPGRADE_ELASTICSEARCH
```
antonym: Because this prompts for user input, this will break gating scripts.  The openstack-ansible-ops has a VALIDATE_UPGRADE_INPUT which lets you bypass the prompts but that's not at this lower level.  I'd rather we just set default vars for those two here, and allow the user to override if they don't want those done.

Frank-ZhangXin:
I added several lines in pre_leap.sh and post_leap.sh to help automating ES upgrade functions.

scripts/pre_leap.sh
Frank-ZhangXin: Cool, just export vars to execute es-upgrade, and let user override them if they don't like.

Frank-ZhangXin:
@antonym changes are applied.

scripts/pre_leap.sh
```diff
@@ -19,15 +19,10 @@
...
-    echo 'Elasticsearch logs will be preserved.'
-    export UPGRADE_ELASTICSEARCH="YES"
-    export CONTAINERS_TO_DESTROY='all_containers:!galera_all:!neutron_agent:!ceph_all:!rsyslog_all:!elasticsearch_all'
-else
-    echo 'Elasticsearch logs will not be preserved.'
-fi
+# Preserve Elasticsearch data in the leap
+echo 'Elasticsearch logs will be preserved.'
+export UPGRADE_ELASTICSEARCH="YES"
+export CONTAINERS_TO_DESTROY='all_containers:!galera_all:!neutron_agent:!ceph_all:!rsyslog_all:!elasticsearch_all'
```
antonym: You'll probably want something like this so they can be overridden:
    export UPGRADE_ELASTICSEARCH="${UPGRADE_ELASTICSEARCH:-YES}"
    export CONTAINERS_TO_DESTROY='"'"${CONTAINERS_TO_DESTROY:-all_containers:!galera_all:!neutron_agent:!ceph_all:!rsyslog_all:!elasticsearch_all}"'"'

Frank-ZhangXin:
@antonym export has been fixed

scripts/pre_leap.sh
```diff
@@ -20,9 +20,8 @@ set -e -u
 set -o pipefail
 
 # Preserve Elasticsearch data in the leap
-echo 'Elasticsearch logs will be preserved.'
-export UPGRADE_ELASTICSEARCH="YES"
-export CONTAINERS_TO_DESTROY='all_containers:!galera_all:!neutron_agent:!ceph_all:!rsyslog_all:!elasticsearch_all'
+export UPGRADE_ELASTICSEARCH={$UPGRADE_ELASTICSEARCH:-"yes"}
+export CONTAINERS_TO_DESTROY={$CONTAINERS_TO_DESTROY:-'all_containers:!galera_all:!neutron_agent:!ceph_all:!rsyslog_all:!elasticsearch_all'}
```
antonym: Should look like these, dollar sign should be outside the bracket and probably need the quotes set like we did upstream to prevent bash from executing the !s:

export UPGRADE_ELASTICSEARCH="${UPGRADE_ELASTICSEARCH:-YES}"
export CONTAINERS_TO_DESTROY='"'"${CONTAINERS_TO_DESTROY:-all_containers:!galera_all:!neutron_agent:!ceph_all:!rsyslog_all:!elasticsearch_all}"'"'

Frank-ZhangXin: Cool, will correct them.

scripts/pre_leap.sh
```diff
@@ -19,6 +19,10 @@
 set -e -u
 set -o pipefail
 
+# Preserve Elasticsearch data in the leap
+export UPGRADE_ELASTICSEARCH=${UPGRADE_ELASTICSEARCH:-"yes"}
+export CONTAINERS_TO_DESTROY=${CONTAINERS_TO_DESTROY:-""all_containers:!galera_all:!neutron_agent:!ceph_all:!rsyslog_all:!elasticsearch_all""}
```
cloudnull: No need to have the `""` around our default. This should likely just be single quoted. 

``` shell
export CONTAINERS_TO_DESTROY=${CONTAINERS_TO_DESTROY:-'all_containers:!galera_all:!neutron_agent:!ceph_all:!rsyslog_all:!elasticsearch_all'}
```

Additionally I'm not sure we need to have this be configurable? We know these are the container groups and types we want to keep so it may be better for us to just set it and forget it.

``` shell
export CONTAINERS_TO_DESTROY='all_containers:!galera_all:!neutron_agent:!ceph_all:!rsyslog_all:!elasticsearch_all'
```

scripts/elasticsearch_upgrade.sh
```diff
@@ -0,0 +1,39 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+## Shell Opts ----------------------------------------------------------------
+set -e -u -x
+set -o pipefail
+
+# Check upgrade flag
+if ! [[ "$UPGRADE_ELASTICSEARCH" == "YES" ]]; then
```
cloudnull: This should not be here. The only entry point for this script is through the `post-leap.sh` script. If this script is called it should just run.

scripts/post_leap.sh
```diff
@@ -35,4 +35,11 @@ if [[ ! -f "${UPGRADE_LEAP_MARKER_FOLDER}/deploy-rpc.complete" ]]; then
 else
   log "deploy-rpc" "skipped"
 fi
+
+if [ -f "$UPGRADE_ELASTICSEARCH" ]; then
```
cloudnull: This condition is looking for a file is that what we want? It seems better if we have this be a Boolean. Something like:

``` shell
export UPGRADE_ELASTICSEARCH=false


if [ "${UPGRADE_ELASTICSEARCH}" = true ]; then
    pushd /opt/rpc-upgrades/scripts
        source ./elasticsearch_upgrade.sh
    popd
fi
```

As a Boolean we'd not need to worry about string casting and cases.

scripts/pre_leap.sh
antonym: Thought we might have the situation where the customer doesn't want to keep the data and there may not be a point in keeping it.  In this case, they'd have the option to override.

Frank-ZhangXin: Double-quote sometimes lose function and failed to exporting value to 'CONTAINERS_TO_DESTROY', so double them to make sure it will work. I'm testing single-quote and double double-quote now.

antonym:
recheck_continuous-integration/travis-ci/push

antonym:
recheck_all

scripts/elasticsearch_upgrade.sh
```diff
@@ -0,0 +1,33 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+## Shell Opts ----------------------------------------------------------------
+set -e -u -x
+set -o pipefail
+
+pushd /opt/rpc-openstack/rpcd/playbooks
```
antonym: pushd /opt/rpc-upgrades/playbooks since you are running most playbooks from there.  Then you can drop the paths of the files below.

scripts/elasticsearch_upgrade.sh
```diff
@@ -0,0 +1,33 @@
...
+    # Update pip.conf on Elasticsearch container
+    openstack-ansible /opt/rpc-upgrades/playbooks/elasticsearch-postleap-pip-upgrade.yml
+
+    # Stop logstash service on Logstash container
+    openstack-ansible /opt/rpc-upgrades/playbooks/logstash-stop.yml
+
+    # Run elasticsearch upgrade playbook
+    openstack-ansible -e 'logging_upgrade=true' --tags 'reindex-wrapper,elasticsearch-upgrade' elasticsearch.yml
```
antonym: pushd /opt/rpc-openstack/rpcd/playbooks to run elasticsearch.yml

scripts/elasticsearch_upgrade.sh
```diff
@@ -0,0 +1,33 @@
...
+
+    # Restore logstash service on Logstash container
+    openstack-ansible /opt/rpc-upgrades/playbooks/logstash-start.yml
```
antonym: Change back to rpc-upgrades.

scripts/post_leap.sh
```diff
@@ -35,4 +35,11 @@ if [[ ! -f "${UPGRADE_LEAP_MARKER_FOLDER}/deploy-rpc.complete" ]]; then
 else
   log "deploy-rpc" "skipped"
 fi
+
+if [ "$UPGRADE_ELASTICSEARCH"="yes" ]; then
```
antonym: if [ "$UPGRADE_ELASTICSEARCH" == "YES" ];

scripts/elasticsearch_upgrade.sh
```diff
@@ -0,0 +1,33 @@
...
+## Shell Opts ----------------------------------------------------------------
+set -e -u -x
+set -o pipefail
+
+pushd /opt/rpc-openstack/rpcd/playbooks
+    # Update pip.conf on Elasticsearch container
+    openstack-ansible /opt/rpc-upgrades/playbooks/elasticsearch-postleap-pip-upgrade.yml
+
+    # Stop logstash service on Logstash container
+    openstack-ansible /opt/rpc-upgrades/playbooks/logstash-stop.yml
```
BjoernT: Why using a playbook if you can just use ansible to stop the service ?

scripts/elasticsearch_upgrade.sh
BjoernT: Same as above

scripts/post_leap.sh
```diff
@@ -35,4 +35,11 @@ if [[ ! -f "${UPGRADE_LEAP_MARKER_FOLDER}/deploy-rpc.complete" ]]; then
...
+    pushd /opt/rpc-upgrades/scripts
+        source ./elasticsearch_upgrade.sh
```
BjoernT: This operation needs to be backgrounded as the ES index updates take upwards of 24h on normal deployments and we need to "finish" the main upgrade earlier

scripts/elasticsearch_upgrade.sh
Frank-ZhangXin: Cool, will change

scripts/elasticsearch_upgrade.sh
Frank-ZhangXin: OK

scripts/elasticsearch_upgrade.sh
Frank-ZhangXin: OK

scripts/post_leap.sh
Frank-ZhangXin: get it, will fix the equals. But 'yes' is made on purpose to pair with upstream format: https://github.com/rcbops/rpc-openstack/blob/newton/scripts/functions.sh#L62 all vars are exported in lower case.

antonym: yeah, that's fine, mainly was mentioning the double =

scripts/elasticsearch_upgrade.sh
Frank-ZhangXin: Using ansible ad-hoc can achieve same goal, but these scripts were already passed tests. So optimization could be addressed later. For now top goal is merging this functionality.

scripts/elasticsearch_upgrade.sh
Frank-ZhangXin: same to above

scripts/post_leap.sh
Frank-ZhangXin: Yeah, you're right about time consuming of ES reindex. And our original object is reduce interaction from user as less as possible, so automating this part as default. User can override es ugprade/redinex process at the beginning of leap now. Automated part is not mandatory.

Frank-ZhangXin:
@cloudnull @antonym @BjoernT updated commit, pls check replied comment also.

scripts/elasticsearch_upgrade.sh
Frank-ZhangXin: fixed.

scripts/post_leap.sh
Frank-ZhangXin: fixed.

scripts/post_leap.sh
```diff
@@ -35,4 +35,11 @@ if [[ ! -f "${UPGRADE_LEAP_MARKER_FOLDER}/deploy-rpc.complete" ]]; then
 else
   log "deploy-rpc" "skipped"
 fi
+
+if [ "$UPGRADE_ELASTICSEARCH"=="yes" ]; then
```
antonym: Add a space to both sides of the ==.  Other than that LGTM.

scripts/post_leap.sh
```diff
@@ -35,4 +35,11 @@ if [[ ! -f "${UPGRADE_LEAP_MARKER_FOLDER}/deploy-rpc.complete" ]]; then
...
+    pushd /opt/rpc-upgrades/scripts
+        source ./elasticsearch_upgrade.sh
```
BjoernT: Just ruin this script in a screen/tmux and wrap some logic around a lock file, similar to https://github.com/rsoprivatecloud/openstack-ops/blob/master/playbooks/update-hp-firmware.yml#L43-L57 so we can background the task

scripts/post_leap.sh
Frank-ZhangXin: You're right, spaces are better.

scripts/post_leap.sh
Frank-ZhangXin: What about adding 'fire and forget' to this task? So main leap can finish, then user could check reindex by polling es status.
Like
- name: reindex
   command: xxx
   async: 45
   poll: 0

cloudnull:
@Frank-ZhangXin this is failing the lint checks but otherwise looks good to me.

scripts/post_leap.sh
Frank-ZhangXin: IMO, based on current script, when leap progress reaches ES upagrade part, whole leaping has arrived the very end, so there's no other upgrades wait to do. And also user can simply switch out tmux session then polling ES status to check reindexing progress. We made this guide into document too.

BjoernT: I still prefer to background the task and especially we need to increase the ansible command timeout, can't remember which setting this was. Otherwise the thread gets interrupted prematurely. Hence I prefer to nohup the task no matter if it runs in a tmux and at the end of a script

PR #37, opened 2017-11-27
RLM-518 Set maas_swift_accesscheck_password

antonym:
Sets maas_swift_accesscheck_password on older
versions of liberty before deploying in order
for maas-setup to complete.

### DIFF
--- a/tests/aio-create.sh
+++ b/tests/aio-create.sh
@@ -177,6 +177,13 @@ function fix_galera_apt_cache {
   popd
 }
 
+function maas_tweaks {
+  # RLM-518 older versions of liberty did not set maas_swift_accesscheck_password
+  if ! grep '^maas_swift_accesscheck_password\:' /opt/rpc-openstack/rpcd/etc/openstack_deploy/user_extras_secrets.yml; then
+    echo 'maas_swift_accesscheck_password:' >> /opt/rpc-openstack/rpcd/etc/openstack_deploy/user_extras_secrets.yml
+  fi
+}
+
 ## Main ----------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


antonym:
Because liberty branch contains the password already, testing this branch using this periodic:

https://rpc.jenkins.cit.rackspace.net/view/Upgrades/job/PM_rpc-upgrades-master-trusty_aio-swift-r12.2.2_to_newton_leap/20/

tests/aio-create.sh
```diff
@@ -177,6 +177,13 @@ function fix_galera_apt_cache {
   popd
 }
 
+function maas_tweaks {
```
cloudnull: is this something we should simply check for and add into the `user_extras_secrets.yml` file as we're upgrading?

antonym: It could probably be a good item to put in the preflight checks before a leapfrog, but in this case, this breaks before we run leapfrog.  I'm guessing that existing environments that have been upgraded to rpc-maas would have ran into this and probably already had the password deployed.

antonym:
Looks like https://rpc.jenkins.cit.rackspace.net/view/Upgrades/job/PM_rpc-upgrades-master-trusty_aio-swift-r12.2.2_to_newton_leap/20/ is now working correctly.

tests/aio-create.sh
antonym: Created https://rpc-openstack.atlassian.net/browse/RLM-529 to create a pre-flight check.

antonym:
Test passed, merging this so we can test jobs overnight.

PR #41, opened 2017-11-28
RLM-554 Various MaaS fixes

antonym:
Restores playbook to remove old neutron container maas entries
which was removed in RLM-416.

Runs maas-install before and after leapfrog which gets rpc-maas
if it's not present, installs, and verifies.  It also handles
the jumping around of the maas_release variable and sets it to
master by default for now.

Also disables the Elastic Search Upgrade jobs on Kilo and Liberty.  The job currently works on L->N but not L-N-14.5 because the required rpc-o commit won't be in the release until 14.6.0.  It also is not currently supported on Kilo.  Leaving them disabled for now and we can reenable once 14.6.0 drops.

### DIFF
--- a/tests/maas-install.sh
+++ b/tests/maas-install.sh
@@ -18,10 +18,26 @@
 
 set -evu
 
+## Vars ----------------------------------------------------------------------
+export RE_JOB_SERIES="${RE_JOB_SERIES:-newton}"
+
+# sets the desired maas_release to test
+MAAS_RELEASE=master
+
 pushd /opt/rpc-upgrades/playbooks
-  # install maas
-  openstack-ansible maas-get.yml -vv
-  openstack-ansible /opt/rpc-maas/playbooks/site.yml -vv
-  # verify maax is running
+  # checkout rpc-maas
+  if [ ! -d "/opt/rpc-maas" ]; then
+    openstack-ansible maas-get.yml -e maas_release="${MAAS_RELEASE}" -vv
+  fi
+
+  # install rpc-maas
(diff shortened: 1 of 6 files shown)


tests/maas-install.sh
```diff
@@ -18,10 +18,29 @@
 
 set -evu
 
+# sets the desired maas_release to test
+MAAS_RELEASE=master
```
cloudnull: will this ever need to be configurable?

tests/maas-install.sh
```diff
@@ -18,10 +18,29 @@
...
+
+# Ensure we have one copy of the maas_release variable since
```
cloudnull: maybe this would just be easier if we passed the needed variable into the required playbooks? 

``` shell
openstack-anisble maas-get.yml -e maas_release="${MAAS_RELEASE}"
```

tests/maas-install.sh
antonym: Guess it depends on if we stabilize on a version or not... I could do what rpc-o is doing and grab the latest version and test that, but currently the latest version isn't installing yaml which is needed to pass when master does.

tests/maas-install.sh
antonym: Yeah... that probably would be easier... I'll make the change to use that instead.

antonym:
recheck_all

antonym:
recheck_all

antonym:
Recheck_cit_trusty_aio_swift_kilo_to_newton_leap

antonym:
Recheck_cit_trusty_aio_swift_kilo_to_newton_leap

PR #43, opened 2017-11-29
Use latest OA_OPS for ES and don't run on Kilo

antonym:
Uses the latest OA_OPS to pull in the containers_to_destroy work, doesn't upgrade ES on kilo, and fixes a playbook syntax issue in the reindex playbook.


antonym:
Passes on newton, so merging in to fix gates.

PR #45, opened 2017-11-30
RLM-434 Implement Ansible Retries for older ver

antonym:
Switches to ansible release with retries added
to alleviate random ssh timeouts that plague
leapfrog runs from passing occasionally.

### DIFF
--- a/tests/aio-create.sh
+++ b/tests/aio-create.sh
@@ -183,6 +183,13 @@ function maas_tweaks {
   fi
 }
 
+function ansible_retry {
+  # RLM-434 Implement ansible retries for mitaka and below
+  export ANSIBLE_SSH_RETRIES=3
+  export ANSIBLE_GIT_RELEASE=ssh_retry
+  export ANSIBLE_GIT_REPO=https://github.com/hughsaunders/ansible
+}
+
 ## Main ----------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


tests/aio-create.sh
```diff
@@ -184,6 +184,13 @@ function maas_tweaks {
   fi
 }
 
+function ansible_retry {
+  # RLM-434 Implement ansible retries for mitaka and below
+  export ANSIBLE_SSH_RETRIES=3
+  export ANSIBLE_GIT_RELEASE=ssh_retry
+  export ANSIBLE_GIT_REPO=https://github.com/hughsaunders/ansible
```
Frank-ZhangXin: Using Hugh's repo is work around, isn't it?

antonym: Yeah, it's pulling 1.9.4-1 of Ansible with this commit on top of it:

https://github.com/hughsaunders/ansible/commit/4a7c0003501d8c76569ab2efcc8e55261fd8bc9b

It might be better later to maybe have an ansible repo dropped into rcbops with that branch just it's in one location, but mainly just want to see if this improves the random ssh issues the jobs have.

antonym:
Both tests passed, so merging to get some usage over the weekend.

PR #47, opened 2017-11-30
RLM-299 Enable the security role

antonym:
Enables the security role.


antonym:

antonym:
Role already runs during leap, this sets enables it to run during deploy too.

PR #48, opened 2017-12-04
RLM-653 Always attempt to check out the release

antonym:
rpc-deploy-playbooks will checkout rpc-maas even if the
deploy flag is off, so checking it out each pass here
ensures the version we want to deploy with is the one
from this file.


antonym:
recheck_all

antonym:
Tests all pass, merging.

PR #51, opened 2017-12-05
RLM-682 Pin python-ldap 2.5.2 for keystone

antonym:
Pins to the last known working version of python-ldap.


antonym:
Got blessing from @cloudnull too so merging.

PR #52, opened 2017-12-05
RLM-274 Get latest MAAS tag and set as var

antonym:
Retrieves the latest MAAS tag during post leap, and sets the version before deploy-rpc-playbooks runs.  This ensures the latest MAAS version is set and prevents a revert to 1.0.0 as is set in 14.2.0 release.


antonym:

PR #54, opened 2017-12-08
RLM-735 Install httplib2 for uri check in preflight

antonym:
Pre Ansible 2.1 requires httplib2 to be install for uri
module.


antonym:
Job shows successful install of package, merging.

PR #70, opened 2017-12-15
RLM-900 Set the terminal before running Leapfrog

antonym:
Jenkins doesn't seem to be happy if the term is not set and tosses
a 'unknown': I need something more specific. when running leapfrog.

This attempts to get around that issue.


antonym:
Liberty to Newton has gotten past the issue now and is running leapfrog.

antonym:

PR #71, opened 2017-12-15
RLM-899 Ensure ansible retries are getting ran for AIO and MNAIO

antonym:
Previously retries for AIO was not getting sourced any more
since breaking out prepare-rpco and create-aio.  This should
handle it better and cleans up the code a bit.

### DIFF
--- a/tests/create-aio.sh
+++ b/tests/create-aio.sh
@@ -35,6 +35,15 @@ if [ ! -d "/opt/rpc-openstack" ]; then
   popd
 fi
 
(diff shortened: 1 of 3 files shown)


antonym:
Not really, but 1.9s retry ability was pretty poor, stability has improved quite a bit using the fork so I'm inclined to use it for the time being.  Maybe we fork ansible into rcbops and move the branch there later on if we're concerned about it living in someones individual repo.

PR #75, opened 2018-01-02
RLM-292 Add second leapfrog run to gating

antonym:
To ensure leapfrog is idempotent.

PR #76, opened 2018-01-02
 RLM-293 Secondary qc run after second pass of leap

antonym:
Merge https://github.com/rcbops/rpc-upgrades/pull/75 first


antonym:
Moved this to https://github.com/rcbops/rpc-upgrades/pull/75 so closing.

PR #82, opened 2018-01-09
RLM-1088 Add docs for running tests on an MNAIO

antonym:
Adds docs for testing an MNAIO leapfrog using gating scripts.

### DIFF
--- a/testing.rst
+++ b/testing.rst
@@ -14,10 +14,17 @@ desired branch, set the RE_JOB_ACTION and run the script
 `run-tests.sh`. This will perform the upgrade exactly as gating
 job hooks will.
 
+It should follow this format:
+
 .. code-block:: shell
 
     RE_JOB_ACTION=<from_version>_to_<to_version>_<upgrade_action>
-    RE_JOB_ACTION=kilo_to_newton_leap ./run-tests.sh
+
+An example would be:
+
+.. code-block:: shell
+
+    export RE_JOB_ACTION=kilo_to_newton_leap ./run-tests.sh
 
 
 If you wish to run the tests against a specific version, set the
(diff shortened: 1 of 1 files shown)


testing.rst
```diff
@@ -26,10 +33,36 @@ you wish to test with.
...
+Build an OnMetal server using the flavor onmetal-io1 and running
+Ubuntu 16.04 LTS.
+
+You'll need to set the RE_JOB_IMAGE to xenial_mnaio in order to
+trigger the mnaio code path.  Then set the action to set the
+leap you want to test.
+
+.. code-block:: shell
+
+    export RE_JOB_IMAGE=xenial_mnaio
```
claco: Does this tell o-a-ops to spin it's vms as xenial vms?
If so, shouldn't this be trusy_mnaio?
If not, maybe some clarifying explanation.

antonym: Yeah, my bad... I've been in xenial mode this week, I'll correct.

antonym: Eh, actually that's correct... we're using a Xenial Host now for MNAIO testing, the image that it pulls down will be Trusty though.

antonym: This is where you can tune the gating script to use Trusty or Xenial as the VM.  Right now the assumption is all leapfrogs start from Trusty: 

https://github.com/rcbops/rpc-upgrades/blob/master/tests/create-mnaio.sh#L39

claco: Cool. Thanks for the clarification. We might want to just dump that comment/blurb/knowledge into this doc so the next person reading it knows that too.

antonym: Added a quick blurb about it.

PR #84, opened 2018-01-11
RLM-322 Fix domain naming

Frank-ZhangXin:
Since current naming method in ansible-ansible-openstack_hosts role
would cause unexpected extra peroid sign in the name. So we remove
the perioid sign appeared in the function 'host_update' of
openstack_hosts role. Then make openstack_domain default as empty
string.


Frank-ZhangXin:

claco:
Double checking that this also fixes the issue where the customer or service deliver sets the domain to blank/empty "", leaving us with two hostnames in /etc/hosts:

```
x.x.x.x.   hostname. hostname
```

And this is what causes the service shenanigans in the database.

scripts/fix_domain.sh
```diff
@@ -0,0 +1,32 @@
...
+
+# Fix peroid sign on 'host_update' function
+sed -i 's/${RFCHOSTNAME}.${DOMAINNAME}/${RFCHOSTNAME}${DOMAINNAME}/g' /etc\
+\ansible\roles\openstack_hosts/templates/openstack-host-hostfile-setup.sh.j2
+
+# Fix 'openstack_host' in OSA-ops leap-upgrade
+sed -i 's/openstack.local/.openstack.local/g' /opt/rpc-leapfrog\
+/openstack-ansible-ops/leap-upgrades/upgrade-utilities-mitaka/playbooks\
+/old-hostname-compatibility-mitaka.yml
+sed -i 's/openstack.local/.openstack.local/g' /opt/rpc-leapfrog\
```
antonym: We probably should do a check to see if these conditions exist before running a sed.  I could see this sed causing a problem if ran multiple times (which the script is supposed to be able to do) as I think it would add a dot per run.

Frank-ZhangXin:
@claco yes, it's supposed to fix it

scripts/fix_domain.sh
Frank-ZhangXin: yeah, good ideas.

Frank-ZhangXin:
Upstream OA-ops fix PR: https://review.openstack.org/533759 RLM-322 Fix domain name

Frank-ZhangXin:
@antonym I updated it, fix hardcoded 'dot' sign with bash script then sent one OA-ops upstream PR to fix upgrade-utility: https://review.openstack.org/#/c/533759/

antonym:
Is this something that we should change upstream in openstack-host-hostfile-setup.sh.j2 instead of within our repo?  Future deployments for minor upgrades would probably break this functionality if the template isn't modified.

Frank-ZhangXin:
@antonym yeah, you're right and agree with you, but I mentioned changing on upstream on Monday, no one responded if we should. For this one looks just like hotfix rather than forever fix.
And BTW the 'sed' always failed looking for 'openstack_hosts' role dir, and I found liberty /etc/ansible/roles doesn't have one in fact. Wonder inserting point of this 'sed' replacement if we really want to do it like temporary fix.

PR #85, opened 2018-01-15
[RLM-712] Clean glance cache before leap

Frank-ZhangXin:
This action will wipe all glance image cache before leaping to prevent
the container too full to achieve upgrading.

### DIFF
--- a/playbooks/glance-cache-cleanup.yml
+++ b/playbooks/glance-cache-cleanup.yml
@@ -43,10 +43,10 @@
       service:
         name: glance-registry
         state: started
-        patter: glance-registry
+        pattern: glance-registry
 
     - name: Restore glance-api
       service:
         name: glance-api
         state: started
-        patter: glance-api
+        pattern: glance-api
--- a/playbooks/glance-cache-cleanup.yml
+++ b/playbooks/glance-cache-cleanup.yml
@@ -1,0 +1,52 @@
+---
+# Copyright 2017, Rackspace US, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+- name: Stop glance services
+  hosts: glance_all
+  user: root
+  tasks:
+    - name: Stop glance-api
+      service:
+        name: glance-api
+        state: stopped
+        pattern: glance-api
+
+    - name: Stop glance-registry
+      service:
+        name: glance-registry
+        state: stopped
+        pattern: glance-registry
+
+    - name: Wipe glance cache directory
+      file:
+        state: absent
+        path: "/var/lib/glance/cache/"
+
+    - name: Generate cache directory
+      file:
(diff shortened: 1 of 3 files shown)


playbooks/glance-cache-cleanup.yml
```diff
@@ -0,0 +1,52 @@
...
+      file:
+        state: absent
+        path: "/var/lib/glance/cache/"
+
+    - name: Generate cache directory
+      file:
+        state: directory
+        path: "/var/lib/glance/cache/"
+
+    - name: Restore glance-api
```
antonym: Might be a good idea to start glance-registry back up first since it's the datastore API talks to.

antonym:
Looks like gate is failing with: ERROR: /opt/rpc-openstack/openstack-ansible/playbooks/inventory/env.d/aodh.yml:5: Invalid ini entry: aodh_all - need more than 1 value to unpack

Usually this means the inventory doesn't match the proper version of ansible.  Maybe this needs to run earlier before anything is modified or upgraded.

playbooks/glance-cache-cleanup.yml
Frank-ZhangXin: cool

Frank-ZhangXin:
@antonym I moved timing of glance cache wipe to beginning of pre_leap script, hope it's soon enough to avoid aodh error on gate

antonym:
I think it needs to run before this point here:

https://github.com/rcbops/rpc-upgrades/blob/master/scripts/ubuntu14-leapfrog.sh#L98

Because after rpc-openstack of the target version is checked out, it contains the inventory for Newton but is still running an older version of Ansible (1.9 or so) that does not understand how to parse the inventory.  

The other option would be to checkout rpc-openstack of the current version you are leaping from so that it can process the inventory right and then move it back to the target.

Frank-ZhangXin:
@antonym I moved the playbook into ubuntu14, hope it would work.

Frank-ZhangXin:
Still got error on gate:
ERROR: /opt/rpc-openstack/openstack-ansible/playbooks/inventory/env.d/aodh.yml:5: Invalid ini entry: aodh_all - need more than 1 value to unpack

scripts/ubuntu14-leapfrog.sh
```diff
@@ -85,6 +85,11 @@ else
   log "rpc-preflight-check" "skipped"
 fi
 
+# Glance cache cleanup
+pushd /opt/rpc-upgrades/playbooks/
+    openstack-ansible glance-cache-cleanup.yml
+pop
```
antonym: Resolved the conflict, but this should be popd

Frank-ZhangXin: Yeah, fixing it now

antonym:

playbooks/glance-cache-cleanup.yml
```diff
@@ -0,0 +1,52 @@
...
+    - name: Generate cache directory
+      file:
+        state: directory
+        path: "/var/lib/glance/cache/"
+
+    - name: Restore glance-registry
+      service:
+        name: glance-registry
+        state: started
+        patter: glance-registry
```
antonym: Looks like its running in gate now, but two cases where pattern is mispelled is stopping it from running.

playbooks/glance-cache-cleanup.yml
```diff
@@ -0,0 +1,52 @@
...
+
+    - name: Restore glance-api
+      service:
+        name: glance-api
+        state: started
+        patter: glance-api
```
antonym: pattern (sp)

antonym:
Passing in Liberty now, will merge once kilo is good.

antonym:

PR #86, opened 2018-01-15
RLM-1118 Don't fail if rpc-openstack is not present

antonym:
Moves existing rpc-openstack to backup directory if
backup directory doesnt exist.  Checks out rpc-openstack
fresh if not present.  This allows rpc-upgrades to be the
only thing needed to checkout to an enviroment and will
pull the items that are needed.

### DIFF
--- a/scripts/ubuntu14-leapfrog.sh
+++ b/scripts/ubuntu14-leapfrog.sh
@@ -97,8 +97,12 @@ pushd ${LEAPFROG_DIR}
     # Prepare rpc folder
     if [[ ! -f "${UPGRADE_LEAP_MARKER_FOLDER}/rpc-prep.complete" ]]; then
         if [[ ! -d "${LEAPFROG_DIR}/rpc-openstack.pre-newton" ]]; then
-            # Cleanup existing RPC, replace with new RPC
-            mv ${RPCO_DEFAULT_FOLDER} ${LEAPFROG_DIR}/rpc-openstack.pre-newton
+            # if existing RPCO folder exists, back it up
+            if [[ -d ${RPCO_DEFAULT_FOLDER} ]]; then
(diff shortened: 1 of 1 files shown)


scripts/ubuntu14-leapfrog.sh
```diff
@@ -97,8 +97,12 @@ pushd ${LEAPFROG_DIR}
     # Prepare rpc folder
     if [[ ! -f "${UPGRADE_LEAP_MARKER_FOLDER}/rpc-prep.complete" ]]; then
         if [[ ! -d "${LEAPFROG_DIR}/rpc-openstack.pre-newton" ]]; then
-            # Cleanup existing RPC, replace with new RPC
-            mv ${RPCO_DEFAULT_FOLDER} ${LEAPFROG_DIR}/rpc-openstack.pre-newton
+            # if existing RPCO folder exists, back it up
+            if [[ -d ${RPCO_DEFAULT_FOLDER} ]]; then 
+                 mv ${RPCO_DEFAULT_FOLDER} ${LEAPFROG_DIR}/rpc-openstack.pre-newton
```
BjoernT: I would either add -f or add a $$ to make it more idempotent on multiple retries.

antonym: If "${LEAPFROG_DIR}/rpc-openstack.pre-newton" exists though, it shouldn't attempt to run the backup again.  I figure the first copy is the one you really want to save, and it should skip the backup entirely on additional runs and attempt to create the rpc-o directory if it doesn't exist.

claco: @BjoernT I see a new git push and the review is marked outdated. Can you weigh in on this one so we can move it forward? Thanks!

PR #89, opened 2018-01-16
Prevent nesting of es-reindex session

BjoernT:
TMUX does not allow nesting of sessions when running inside a existing
TMUX process. New es-reindex sessions are force created outside of the
current TMUX in order to prevent nesting.
Additionally a check is added to skip the reindex of Elasticsearch
once a es-reindex session is already present.

Closes-Bug: RLM-1127


antonym:
Merging this in so we can test this in the gates with UPGRADE_ELASTICSEARCH.

PR #94, opened 2018-01-19
[RLM-322]Remove hardcoded period mark in hostname

Frank-ZhangXin:
Fix the problem in pre_reploy script and replace original combination
without period mark.

### DIFF
--- a/scripts/pre_redeploy.sh
+++ b/scripts/pre_redeploy.sh
@@ -108,7 +108,7 @@ pushd ${RPCO_DEFAULT_FOLDER}/rpcd/playbooks
     openstack-ansible configure-apt-sources.yml
 popd
 # Fix hostname with fixed period mark
-if [ -f /etc/ansible/roles/openstack_hosts/templates/openstack-host-hostfile-setup.sh.j2 ]; then
+if [[ -f "/etc/ansible/roles/openstack_hosts/templates/openstack-host-hostfile-setup.sh.j2" ]]; then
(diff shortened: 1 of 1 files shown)


scripts/pre_redeploy.sh
```diff
@@ -107,3 +107,11 @@ popd
 pushd ${RPCO_DEFAULT_FOLDER}/rpcd/playbooks
     openstack-ansible configure-apt-sources.yml
 popd
+# Fix hostname with fixed period mark
+if [ -f /etc/ansible/roles/openstack_hosts/templates/openstack-host-hostfile-setup.sh.j2 ]; then
```
antonym: Minor nit, wrap in double brackets and put file name in quotes to match reset of formatting in file.

rexredinger: @antonym Is the request for double brackets coming from a particular stylistic preference or just defensive practice?  I'm asking because it touches on an outstanding question I have.  Do we, as a team, have any consistent style guides or best practices defined (formally or informally)?  It's not a huge deal, but when nits like this come up, it's really nice to be able to point to documentation.

antonym: Mainly it came from the rest of the file was using if [[ ]] checks so I mentioned it more for consistency.  The quotes jumped more out at me.  I don't think there's really any formal guidelines in place at this time.

Frank-ZhangXin:
We conducted a new way to override nova.conf to achieve fixing nova hostname.

PR #95, opened 2018-01-22
Adding libev4 for liberty xtrabackup patch

BjoernT:
The new xtrabackup requires libev4 but as we are
downloading and installing xtrabackup via dpkg, dependencies are
not resolved. The pre installation of libev4 before xtrabackup
will fix the missing dependency


antonym:
Will merge this and kick off the failing jobs.

BjoernT:
Thanks we need to make sure that the fix_galera_apt_cache gets executed, otherwise the patch file does not get pulled into the code base

antonym:
Yeah, the gate should by default bring that in on Liberty builds.

PR #99, opened 2018-01-23
RLM-1234 Remove xtrabackup from galera_client

antonym:
xtrabackup is only required for galera_server so
remove it if present in liberty in the
galera_client install.


antonym:
Seems to apply cleanly to 12.1.2, 12.2.2, and 12.2.5 which are the ones that still carry that package in galera_client.

antonym:
Refactored to skip modifying the defaults and just rip out the install actions for the packages.

antonym:
Passing liberty, so merging to fire off in gates.

PR #104, opened 2018-01-24
RLM-271 Remove ansible_ssh from all user configs

BjoernT:
ansible_ssh can be configured in any of the user yml
configuration files, which is mainly set below Newton releases
to roll out MAAS. As it breaks ansible 2.1 we need to remove it
from all occurences rather than just named files.


BjoernT:
@claco What is the requested change ? Just changing the commit message?

PR #105, opened 2018-01-24
RE-1162 Replace Travis CI usage

mattt416:
This commit replaces the Travis CI usage with jobs defined and run
within the standard rpc-gating Jenkins environment.  The environment
variable FUNCTIONAL_TEST is removed and instead we check if
RE_JOB_ACTION is not set to "tox-test" to determine if the functional
test should run.  Otherwise, we run tox w/ the specific environment
passed in via RE_JOB_SCENARIO.

NOTE: The .travis.yml file will be removed in a subsequent review.


mattt416:
recheck_all

PR #106, opened 2018-01-24
[RLM-273]Add duplicate flavor IDs check

Frank-ZhangXin:
Duplicate nova flavor exisiting would cause 'nova-manage db sync' failed
at the mirgration step. This check could prevent this case happened adn
let user have chance to cleanup db before leapupgrade.


antonym:
Can you squash and add JIRA number to commit?

PR #81, opened 2018-01-05
Adds a flag for skipping UPGRADES, default is yes

antonym:
Adds a flag for skipping upgrades in the event a tester wants to do the first part of the deploy but hold off on the remaining steps.

PR #108, opened 2018-01-25
Add var to override running of upgrade in tests

jacobwagner:
Currently we upgrade no matter what the user is trying to in the
tests directory of scripts. This commit adds an override variable
to allow the tester to choose whether to upgrade or not at run time
of the build.sh script. This will allow the user to have more granular
control when testing


jacobwagner:
Closing this in lue of https://github.com/rcbops/rpc-upgrades/pull/81

PR #109, opened 2018-01-25
Split MNAIO Upgrade from MNAIO Create Script

jacobwagner:
This commit pulls the MNAIO upgrade out of the
MNAIO create script and puts it into its own
script so that testers can call the upgrade at
a later time if needed


antonym:
LGTM, gates will noop on this so merging.

PR #110, opened 2018-01-25
RLM-713 Set backup_dir as a fact to ensure consistency

antonym:
Existing variable is dynamic and can change during a run.
Setting to a fact ensures the variable doesn't change during
a run.


BjoernT:
BTW usually we set BKUPDIR as environment variable so it won't change once you surpass midnight

PR #111, opened 2018-01-29
RLM-1291 Adds hook in for major upgrades

antonym:
Currently fires off the test-major.sh script for the major upgrade.

There are a few tweaks that will need to be done around the config
file migrations that will need to be done to get it fully working.


antonym:

PR #113, opened 2018-01-30
Fix bashate error

antonym:
E004: File did not end with a newline


antonym:
recheck_cit_container_ansible-syntax_tox-test

PR #116, opened 2018-01-31
RLM-1291 Add some fixes for Mitaka to Newton

antonym:
Adds a few tweaks to get Mitaka to Newton to roll.


antonym:
Merging to test Mitaka gate.

PR #119, opened 2018-02-01
[RLM-319] Ensure Glance Auth v2 for Cloudfiles

rexredinger:
Cleaned up and retrying

### DIFF
--- a/playbooks/preflight-check.yml
+++ b/playbooks/preflight-check.yml
@@ -87,6 +87,26 @@
   tags:
     - leap-fs-check
 
+- name: Ensure correct Glance auth version
+  hosts: localhost
+  gather_facts: false
+  user: root
+  vars:
+    override_file: '/etc/openstack_deploy/user_osa_variables_overrides.yml'
+    auth_key: 'glance_swift_store_auth_version'
+  tasks:
+    - name: Set glance_swift_auth_version to 2 for CloudFiles
+      lineinfile:
+        dest: "{{ override_file }}"
+        regexp: "^{{ auth_key }}:.*"
+        line: "{{ auth_key }}: 2"
+        state: present 
+      when: 
+        - glance_swift_store_auth_address is defined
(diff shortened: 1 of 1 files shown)


rexredinger:

rexredinger:

antonym:
Looks like the job is failing on:

13:04:03 ERROR: set_fact is not a legal parameter of an Ansible Play

You might need to move the set_fact into the tasks section since I don't think it allows set_fact to work where it presently is.

rexredinger:
Thanks!  must have scanned past that message when searching through `/ERROR` matches.  I don't see any need for the flag to a fact, since it's value is pretty short-lived.  So I just grouped it in with the rest of the vars.

rexredinger:

rexredinger:

antonym:
Give something like this a try:
```
- name: Ensure correct Glance auth version
  hosts: localhost
  gather_facts: false
  user: root
  vars:
    override_file: '/etc/openstack_deploy/user_osa_variables_overrides.yml'
    auth_key: 'glance_swift_store_auth_version'
  tasks:
    - name: Set glance_swift_auth_version to 2 for CloudFiles
      lineinfile:
        dest: "{{ override_file }}"
        regexp: "^{{ auth_key }}:.*"
        line: "{{ auth_key }}: 2"
        state: present
      when:
        - glance_swift_store_auth_address is defined
        - glance_swift_store_auth_address | search("api.rackspacecloud")
  tags:
    - cloudfiles-check
```

I ran into a few snags running the existing version, the match didn't seem to be doing things right and if the variable wasn't defined it would fail.  The line in file also wasn't replacing the variable but adding additional variables each run.  By adding the regexp, it does a replace or add instead and prevents duplicate entries from being ran.  Give it a try and let me know if you have any success with that.

antonym:
You'll probably want to use a regex for this line:

regexp: "^{{ auth_key }}:.*"

That way it'll match the key and replace it with the right value.

PR #120, opened 2018-02-01
RLM-1338 Set pin of setuptools if not set

antonym:
Older versions of kilo that use older setuptools fail
with install_requires must be a string  or list of strings
so use the latest version that will work.  Uses the version in
r11.1.9.


antonym:
Merge to test as this only applies to kilo.

PR #123, opened 2018-02-05
[RLM-322]Set 'openstack_domain' value

Frank-ZhangXin:
Add 'openstack_domain' value setting into pre_redeploy.

### DIFF
--- a/scripts/pre_redeploy.sh
+++ b/scripts/pre_redeploy.sh
@@ -106,8 +106,14 @@ pushd ${LEAPFROG_DIR}
 popd
 
 if [[ ! -f "${UPGRADE_LEAP_MARKER_FOLDER}/configure-apt-sources-rpc.complete" ]]; then
-   pushd ${RPCO_DEFAULT_FOLDER}/rpcd/playbooks
-       openstack-ansible configure-apt-sources.yml
-   popd
-   log "configure-apt-sources-rpc" "ok"
+    pushd ${RPCO_DEFAULT_FOLDER}/rpcd/playbooks
+        openstack-ansible configure-apt-sources.yml
+    popd
+    log "configure-apt-sources-rpc" "ok"
 fi
+
+# Set openstack_domain to empty which makes fqdns hostname correct
+pushd /etc/openstack_deploy/
+    sed -i '/^openstack_domain.*/d' user_osa_variables_defaults.yml
+    echo "openstack_domain: ''" >> user_osa_variables_defaults.yml
+popd


scripts/pre_leap.sh
```diff
@@ -42,3 +42,6 @@ fi
 
 # RLM-682 versions past 2.5.2 break things
 echo "python-ldap==2.5.2" >> /opt/rpc-openstack/openstack-ansible/global-requirement-pins.txt
+
+# RLM-322 set domain name to empty
+export openstack_domain=""
```
antonym: This needs to be an Ansible variable that needs to be injected into the overrides file to be picked up properly.  Probably ping @BjoernT to find out which config file they usually modify so we can standardize on the location to insert the variable.

Frank-ZhangXin:
@BjoernT @JCallicoat 
Could you guys point out what exact value I'm supposed to override?
https://github.com/rcbops/openstack-ansible-openstack_hosts/blob/stable/newton/templates/openstack-host-hostfile-setup.sh.j2#L43
It looks not an env var.

BjoernT:
Thats just a normal override via user_variables.yml like ```openstack_domain: ""```

Frank-ZhangXin:
@BjoernT @JCallicoat

scripts/pre_redeploy.sh
```diff
@@ -106,8 +106,16 @@ pushd ${LEAPFROG_DIR}
...
-       openstack-ansible configure-apt-sources.yml
-   popd
-   log "configure-apt-sources-rpc" "ok"
+    pushd ${RPCO_DEFAULT_FOLDER}/rpcd/playbooks
+        openstack-ansible configure-apt-sources.yml
+    popd
+    log "configure-apt-sources-rpc" "ok"
 fi
+
+pushd /etc/openstack_deploy/
```
antonym: Made a few changes and tested this out, seems to work:

```
pushd /etc/openstack_deploy/
    if ! grep "openstack_domain" user_osa_variables_defaults.yml > /dev/null; then
        echo "openstack_domain: ''" >> user_osa_variables_defaults.yml
    else
        sed -i s/^openstack_domain.*/openstack_domain:\ \'\'/g user_osa_variables_defaults.yml
    fi
popd
```
The existing one had a few typos on the file name and would also add openstack_domain to the end of the file each time it ran.

playbooks/preflight-check.yml
```diff
@@ -118,3 +118,12 @@
     - name: Duplicate flavors exist
       fail: msg="Duplicated flavor IDs are detected."
       when: flavor_check.stdout != ""
+
+- name: Verify openstack_domain
```
antonym: Can probably drop preflight as it'll just cause extra noise in failing and won't really matter since we inject it later on anyways.

scripts/pre_redeploy.sh
```diff
@@ -106,8 +106,17 @@ pushd ${LEAPFROG_DIR}
...
+    log "configure-apt-sources-rpc" "ok"
 fi
+
+# Set openstack_domain to empty which makes fqdns hostname correct
+pushd /etc/openstack_deploy/
+    if ! grep "openstack_domain" user_osa_variables_defaults.yml > /dev/null; then
+        echo "openstack_domain: ''" >> user_osa_variables_default.yml
+    else
+        sed -i s/^openstack_domain.*/openstack_domain:\ \'\'/g user_osa_variables_default.yml
+    fi
```
antonym: Trailing whitespace

Frank-ZhangXin: Fixed

Frank-ZhangXin:
Squashed commits.

scripts/pre_redeploy.sh
```diff
@@ -106,8 +106,17 @@ pushd ${LEAPFROG_DIR}
...
+    pushd ${RPCO_DEFAULT_FOLDER}/rpcd/playbooks
+        openstack-ansible configure-apt-sources.yml
+    popd
+    log "configure-apt-sources-rpc" "ok"
 fi
+
+# Set openstack_domain to empty which makes fqdns hostname correct
+pushd /etc/openstack_deploy/
+    if ! grep "openstack_domain" user_osa_variables_defaults.yml > /dev/null; then
+        echo "openstack_domain: ''" >> user_osa_variables_default.yml
```
antonym: Should be user_osa_variables_defaults.yml

scripts/pre_redeploy.sh
```diff
@@ -106,8 +106,17 @@ pushd ${LEAPFROG_DIR}
...
+    else
+        sed -i s/^openstack_domain.*/openstack_domain:\ \'\'/g user_osa_variables_default.yml
```
antonym: Same, should be user_osa_variables_defaults.yml

Frank-ZhangXin: yeah, so it is.

scripts/pre_redeploy.sh
```diff
@@ -106,8 +106,17 @@ pushd ${LEAPFROG_DIR}
...
+    popd
+    log "configure-apt-sources-rpc" "ok"
 fi
+
+# Set openstack_domain to empty which makes fqdns hostname correct
+pushd /etc/openstack_deploy/
+    if ! grep "openstack_domain" user_osa_variables_defaults.yml > /dev/null; then
+        echo "openstack_domain: ''" >> user_osa_variables_defaults.yml
+    else
+        sed -i s/^openstack_domain.*/openstack_domain:\ \'\'/g user_osa_variables_defaults.yml
```
BjoernT: I would always single quote the sed script
Perhaps you can just simplify it via:

```
sed -i '/^openstack_domain.*/d' user_osa_variables_defaults.yml
echo "openstack_domain: ''" >> user_osa_variables_defaults.yml
```

As you seem to overwrite any opensack_domain settings regardless if existing or not

Frank-ZhangXin:
@BjoernT changed code as you said

PR #125, opened 2018-02-06
RLM-1388 Remove prompt to verify secrets

antonym:
Isn't really used and ends up just stopping the process
waiting from user input.


antonym:
recheck_all

PR #126, opened 2018-02-08
RLM-1368 Use latest release tag for gate jobs

antonym:
Allows us to use statically named gate jobs for releases
so that we don't have to change the gate jobs monthly.

Looks for 14.RELEASE in the leap to part of the configuation
and if set, determines the latest r14. tag from RPC-O.


antonym:
Depends on https://github.com/rcbops/rpc-gating/pull/639

PR #131, opened 2018-02-13
RLM-1401 Set package_state to present instead of latest

antonym:
OSA does many attempts to update apt during the deploy process
which can slow down the deployment on environments with high latency.

Using present should get us a similar result and reduce time for deployment.

Also sets haproxy_package_state to latest as the haproxy configuration file apparently requires newer versions of haproxy to load.

### DIFF
--- a/scripts/pre_redeploy.sh
+++ b/scripts/pre_redeploy.sh
@@ -104,7 +104,7 @@ if [[ ! -f "${UPGRADE_LEAP_MARKER_FOLDER}/configure-apt-sources-rpc.complete" ]]
     log "configure-apt-sources-rpc" "ok"
 fi
 
-# Set openstack_domain to empty which makes fqdns hostname correct
+# RLM-322 Set openstack_domain to empty which makes fqdns hostname correct
 pushd /etc/openstack_deploy/
     sed -i '/^openstack_domain.*/d' user_osa_variables_defaults.yml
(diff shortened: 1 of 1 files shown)


antonym:
recheck_all

antonym:
recheck_all

antonym:
recheck_all

Frank-ZhangXin:
HA_proxy install failed

Frank-ZhangXin:

antonym:
Yeah, it seems to keep failing on HAProxy where it can't read some configs.  I wonder if by not getting the latest package, if it's using a package that's too old for the config being generated.  I'm wondering if this will even work successfully.  We'll let it run once more and analyze the logs a bit and see if this is going to be a problem.

antonym:
Since haproxy was the only thing failing, I updated the commit to set an override to allow haproxy to upgrade to the latest version which seemed to let it pass that error.  A quick check of the PR gate times shows:

L2N completes in 3h10m instead of the usual ~3h29m
K2N completes in 3h34m instead of the usual ~3h49m

That combined with the apt cache timeout may give us a little more time back especially in high latency deployments.

antonym:
recheck_all

antonym:
Has passed PR gates three times now with reduced times, merging so we can let it bake over the weekend.

PR #134, opened 2018-02-14
[RLM-1371]Check space for rootfs backup

Frank-ZhangXin:
Since insufficient space for rootfs backup would lead ugrade fail, this
check is added prior all the ugprade begin.

### DIFF
--- a/playbooks/preflight-check.yml
+++ b/playbooks/preflight-check.yml
@@ -74,18 +74,45 @@
       shell: |
         df -h | tail -n +2 | sed s/%//g | awk '{if($5 > 97) print "Alert: "$5 "% full - "$0}'
       register: fs_check
+      tags:
+        - leap-fs-check
 
     - name: Check if any filesystem is greater than 97% full
       debug:
         msg: >-
           The filesystem(s) found to be (nearly) full.
-          Details: {{ item.split() | join(' ') }}
+          Details: "{{ item.split() | join(' ') }}"
       failed_when: true
       when:
         - fs_check.stdout_lines
       with_items: "{{ fs_check.stdout_lines }}"
-  tags:
-    - leap-fs-check
+      tags:
+        - leap-fs-check
+
+# RLM-1371 detect if enough space for rootfs backup
+- name: Check LXC file system usage for backup
+  hosts: "os-infra_hosts"
+  gather_facts: false
+  user: root
+  tasks:
+    - name: Gather lxc filesystem used percentages
+      shell: |
+        df -h {{ lxc_container_directory | default('/var/lib/lxc') }} | tail -n +2 | sed s/%//g | awk '{if($5 > 50) print "Alert: "$5 "% full - "$0}'
+      register: lxc_fs_check
+      when:
+        - enable_lxc_fs_check | default(true) | bool
+
+    - name: Check if lxc filesystem is greater than 50% full
+      debug:
+        msg: >-
+          The LXC filesystem(s) have been found to be greater than 50% full.
+          Details: "{{ item.split() | join(' ') }}"
+          If you have fully awareness of the potential risk having failure at rootfs backup stage, you can skip this check by passing the option 'enable_lxc_fs_check=false'.
+      failed_when: true
+      when:
+        - lxc_fs_check.stdout_lines
+        - enable_lxc_fs_check | default(true) | bool
+      with_items: "{{ lxc_fs_check.stdout_lines }}"
 
 - name: Ensure correct Glance auth version
   hosts: localhost
@@ -99,11 +126,11 @@
         dest: "{{ item }}"
         regexp: "^{{ auth_key }}:.*"
         line: "{{ auth_key }}: 2"
-        state: present 
+        state: present
       with_first_found:
         - /etc/openstack_deploy/user_osa_variables_overrides.yml
         - /etc/openstack_deploy/user_variables.yml
-      when: 
+      when:
         - glance_swift_store_auth_address is defined
         - glance_swift_store_auth_address | search("api.rackspacecloud")
   tags:


playbooks/preflight-check.yml
```diff
@@ -138,3 +138,18 @@
     - name: Duplicate flavors exist
       fail: msg="Duplicated flavor IDs are detected."
       when: flavor_check.stdout != ""
+
+# RLM-1371 detect if enough space for rootfs backup
+- name: Check space for rootfs backup on infras
+  hosts: "os-infra-hosts"
```
antonym: Matching the wrong group of hosts, needs an underscore instead of a dash: os-infra_hosts

playbooks/preflight-check.yml
```diff
@@ -138,3 +138,18 @@
...
+  tasks:
+    - name: Indentify the space used in /
+      shell: |
+        df -h / | tail -n 1 | awk '{print $5}' | sed 's/%//'
+      register: rootfs_space_used
+
+    - name: Check space if enough
+      fail:
+        msg: "Error: Rootfs backup will fail due to insufficient space left."
+      when: rootfs_space_used > '50'
```
antonym: when: rootfs_space_used.stdout > '50'
```
changed: [liberty] => {"changed": true, "cmd": "df -h / | tail -n 1 | awk '{print $5}' | sed 's/%//'", "delta": "0:00:00.006126", "end": "2018-02-15 23:10:09.611664", "invocation": {"module_args": {"_raw_params": "df -h / | tail -n 1 | awk '{print $5}' | sed 's/%//'", "_uses_shell": true, "chdir": null, "creates": null, "executable": null, "removes": null, "warn": true}, "module_name": "command"}, "rc": 0, "start": "2018-02-15 23:10:09.605538", "stderr": "", "stdout": "61", "stdout_lines": ["61"], "warnings": []}
```

playbooks/preflight-check.yml
```diff
@@ -138,3 +138,18 @@
     - name: Duplicate flavors exist
       fail: msg="Duplicated flavor IDs are detected."
       when: flavor_check.stdout != ""
+
+# RLM-1371 detect if enough space for rootfs backup
+- name: Check space for rootfs backup on infras
+  hosts: "os-infra-hosts"
+  user: root
+  tasks:
+    - name: Indentify the space used in /
```
antonym: nit (sp): Identify

playbooks/preflight-check.yml
```diff
@@ -138,3 +138,18 @@
...
+      shell: |
+        df -h / | tail -n 1 | awk '{print $5}' | sed 's/%//'
+      register: rootfs_space_used
+
+    - name: Check space if enough
```
antonym: Might be good to make the description a bit more verbose so operators know the threshold without having to dive into the code in the event of a failure:

"Check if there is at least 50% space available on Infra nodes"

playbooks/preflight-check.yml
```diff
@@ -138,3 +138,21 @@
     - name: Duplicate flavors exist
       fail: msg="Duplicated flavor IDs are detected."
       when: flavor_check.stdout != ""
+
+# RLM-1371 detect if enough space for rootfs backup
+- name: Check space for rootfs backup on infras
+  hosts: "os-infra_hosts"
+  user: root
+  tasks:
+    - name: Indentify the space used in /
```
antonym: Spelling, should be Identify

playbooks/preflight-check.yml
```diff
@@ -138,3 +138,21 @@
...
+        df -h / | tail -n 1 | awk '{print $5}' | sed 's/%//'
+      register: rootfs_space_used
+
+    - name: Check space if enough for rootfs backup
+      fail:
+        msg: 
+          - "Error: There wasn't at least 50% space available on infra nodes for rootfs backup."
+          - "You can ignore this check by skipping tag 'rootfs_backup_space_check' on 'preflight-check.yml' manually. Then rerun ubuntu14-leapfrog.sh."
+      when: rootfs_space_used.stdout > '50'
+      tags: rootfs_backup_space_check
```
antonym: Can we move this under the main play and make it a list like the others in the file?

    tags:
      - rootfs_backup_space_check

If they skip the tag, they'll probably just want to skip the whole play.

playbooks/preflight-check.yml
```diff
@@ -138,3 +138,21 @@
...
       when: flavor_check.stdout != ""
+
+# RLM-1371 detect if enough space for rootfs backup
+- name: Check space for rootfs backup on infras
+  hosts: "os-infra_hosts"
+  user: root
+  tasks:
+    - name: Indentify the space used in /
+      shell: |
+        df -h / | tail -n 1 | awk '{print $5}' | sed 's/%//'
```
BjoernT: Don't use root use /var/lib/lxc (or a respective override) in case the LXC stuff is on a sub volume

playbooks/preflight-check.yml
BjoernT: Yes just move the tag up to the main play

cloudnull:
:-1: Is this not covered by the following check - [ https://github.com/rcbops/rpc-upgrades/blob/master/playbooks/preflight-check.yml#L68-L88 ]?

Frank-ZhangXin:
@cloudnull IMO they are overlapped a lot but threshold is different here, 50% spare space check is only heads up for operator/user, to make sure they are fully aware about the potential risk upgrade might fail due to backup failure.

cloudnull:
If they are overlapping you can add to the existing implementation and just change the path to something like:

``` shell
 df -h /var/lib/lxc | tail -n +2 | sed s/%//g | awk '{if($5 > 50) print "Alert: "$5 "% full - "$0}'
```

playbooks/preflight-check.yml
```diff
@@ -138,3 +138,22 @@
     - name: Duplicate flavors exist
       fail: msg="Duplicated flavor IDs are detected."
       when: flavor_check.stdout != ""
+
+# RLM-1371 detect if enough space for rootfs backup
+- name: Check space for rootfs backup on infras
+  hosts: "os-infra_hosts"
+  user: root
+  tags:
```
cloudnull: tags at the bottom of the play for consistency.

playbooks/preflight-check.yml
```diff
@@ -138,3 +138,22 @@
...
+    - rootfs_backup_space_check
+  tasks:
```
cloudnull: These tasks can be added to the above to the https://github.com/rcbops/rpc-upgrades/blob/master/playbooks/preflight-check.yml#L68-L88 playbook and limited using a "when" for hosts in the "os-infra_hosts" group.


example:

``` yaml
- name: Gather filesystem used percentages
  shell: |
    df -h | tail -n +2 | sed s/%//g | awk '{if($5 > 50) print "Alert: "$5 "% full - "$0}'
  register: lxc_fs_check
  when:
    - inventory_host in groups["os-infra_hosts"]

- name: Check if any filesystem is greater than 50% full
  debug:
    msg: >-
      The filesystem(s) found to be (nearly) full.
      Details: {{ item.split() | join(' ') }}
  failed_when: true
  when:
    - inventory_host in groups["os-infra_hosts"]
    - lxc_fs_check.stdout_lines
  with_items: "{{ lxc_fs_check.stdout_lines }}"
```

cloudnull: If we're only concerned with the LXC file system we should also do some introspection on the `lxc` volume group if present to ensure it has enough extents to create the container clones.

playbooks/preflight-check.yml
```diff
@@ -87,6 +87,31 @@
   tags:
     - leap-fs-check
 
+- name: Gather lxc filesystem used percentages
```
cloudnull: this needs to be within the playbook. At present it's not properly indented.

playbooks/preflight-check.yml
```diff
@@ -87,6 +87,31 @@
...
+  shell: |
+    df -h {{ lxc_container_directory | default('/var/lib/lxc') }} | tail -n +2 | sed s/%//g | awk '{if($5 > 50) print "Alert: "$5 "% full - "$0}'
+  register: lxc_fs_check
+  when:
+    - inventory_host in groups["os-infra_hosts"]
+  tags:
+    - lxc-fs-check
+
+# RLM-1371 detect if enough space for rootfs backup
+- name: Check if lxc filesystem is greater than 50% full
```
cloudnull: this needs to be within the playbook. At present it's not properly indented.

playbooks/preflight-check.yml
```diff
@@ -87,6 +87,31 @@
...
+  debug:
+    msg: >-
+      The LXC filesystem(s) have been found to be greater than 50% full.
+      Details: {{ item.split() | join(' ') }}
+      If you have fully awareness of the potential risk having failure at rootfs backup stage,
+      you can skip this check with tag 'lxc-fs-check' by rerun 'preflight-check.yml'.
```
cloudnull: Instead of skip-tag make a Boolean option a user can set with `-e` and then add it to the conditions of these two tasks.

``` yaml
- name: Gather lxc filesystem used percentages
  shell: |
    df -h {{ lxc_container_directory | default('/var/lib/lxc') }} | tail -n +2 | sed s/%//g | awk '{if($5 > 50) print "Alert: "$5 "% full - "$0}'
  register: lxc_fs_check
  when:
    - lxc_fs_check | default(false) | bool
    - inventory_host in groups["os-infra_hosts"]

# RLM-1371 detect if enough space for rootfs backup
- name: Check if lxc filesystem is greater than 50% full
  debug:
    msg: >-
      The LXC filesystem(s) have been found to be greater than 50% full.
      Details: {{ item.split() | join(' ') }}
      If you're fully aware of the potential risk, you can skip this check with
      by passing the option 'lxc_fs_check=yes'.
  failed_when: true
  when:
    - lxc_fs_check | default(false) | bool
    - inventory_host in groups["os-infra_hosts"]
    - lxc_fs_check.stdout_lines
  with_items: "{{ lxc_fs_check.stdout_lines }}"
```

Frank-ZhangXin: yeah, that's really a good point. But 'lxc_fs_check' both plays as return value and boolean flag as you gave. It looks a little bit complex and ambiguous, so can we give one separated flag variable to indicate skipping?

Frank-ZhangXin:

Frank-ZhangXin:

Frank-ZhangXin:

antonym:
Looks like it's tossing:
```
fatal: [prumtasltnl-327-5054_heat_engine_container-c81dd423] => error while evaluating conditional: inventory_host in groups["os-infra_hosts"]
```

playbooks/preflight-check.yml
```diff
@@ -84,8 +86,30 @@
...
+    # RLM-1371 detect if enough space for rootfs backup
+    - name: Check if lxc filesystem is greater than 50% full
+      debug:
+        msg: >-
+          The LXC filesystem(s) have been found to be greater than 50% full.
+          Details: {{ item.split() | join(' ') }}
+          If you have fully awareness of the potential risk having failure at rootfs backup stage, you can skip this check by passing the option 'enable_lxc_fs_check=false'.
+      failed_when: true
+      when:
+        - inventory_host in groups["os-infra_hosts"]
```
antonym: think this should be 
```
- inventory_hostname in groups["os-infra_hosts"]
```

playbooks/preflight-check.yml
```diff
@@ -84,8 +86,30 @@
...
-    - leap-fs-check
+      tags:
+        - leap-fs-check
+
+    - name: Gather lxc filesystem used percentages
+      shell: |
+        df -h {{ lxc_container_directory | default('/var/lib/lxc') }} | tail -n +2 | sed s/%//g | awk '{if($5 > 50) print "Alert: "$5 "% full - "$0}'
+      register: lxc_fs_check
+      when:
+        - inventory_host in groups["os-infra_hosts"]
```
antonym: think this should be 
```
- inventory_hostname in groups["os-infra_hosts"]
```

playbooks/preflight-check.yml
```diff
@@ -84,8 +86,30 @@
...
+        - enable_lxc_fs_check | default(true) | bool
+
+    # RLM-1371 detect if enough space for rootfs backup
+    - name: Check if lxc filesystem is greater than 50% full
+      debug:
+        msg: >-
+          The LXC filesystem(s) have been found to be greater than 50% full.
+          Details: {{ item.split() | join(' ') }}
```
antonym: Might need to wrap the brackets in quotes, the syntax checker in VIM seemed to not like this line.

playbooks/preflight-check.yml
```diff
@@ -84,8 +86,30 @@
...
+          If you have fully awareness of the potential risk having failure at rootfs backup stage, you can skip this check by passing the option 'enable_lxc_fs_check=false'.
+      failed_when: true
+      when:
+        - inventory_host in groups["os-infra_hosts"]
+        - lxc_fs_check.stdout_lines
+        - enable_lxc_fs_check | default(true) | bool
+      with_items: "{{ lxc_fs_check.stdout_lines }}"
```
antonym: Got this when running it:

[DEPRECATION WARNING]: Skipping task due to undefined Error, in the future this will be a fatal error.: 'dict object' has no attribute 'stdout_lines'.
This feature will be removed in a future release. Deprecation warnings can be disabled by
setting deprecation_warnings=False in ansible.cfg.

Frank-ZhangXin:
@antonym fixed it. It's supposed to be "inventory_hostname in groups['some_group']"

playbooks/preflight-check.yml
Frank-ZhangXin: Cool, let me fix it really quick

playbooks/preflight-check.yml
Frank-ZhangXin: This one is very strange, and looks like it was triggered just because stdout content is including comma ':' mark. Since it's deprecation warning, do we still need to care about it?

PR #135, opened 2018-02-19
[RLM-320] Ensure Maas Holland enabled

rexredinger:


playbooks/roles/rpc_pre_upgrade/tasks/enable_maas_holland.yml
```diff
@@ -0,0 +1,28 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+- name: Ensure maas_holland_enabled is true
+  - lineinfile:
+      name: "/etc/openstack_deploy/user_variables.yml"
```
Frank-ZhangXin: I'm guessing the proper file should be 'user_rpco_variables_defaults.yml' or 'user_rpco_variables_overrides.yml'.

playbooks/roles/rpc_pre_upgrade/tasks/enable_maas_holland.yml
```diff
@@ -0,0 +1,28 @@
...
+      regexp: "^maas_holland_enabled:"
+      line: "maas_holland_enabled: True"
+      state: present
+
+- name: Ensure maas_holland_venv_enabled is true
+  - lineinfile:
+      name: "/etc/openstack_deploy/user_variables.yml"
```
Frank-ZhangXin: Same, see above comment

playbooks/roles/rpc_pre_upgrade/tasks/enable_maas_holland.yml
rexredinger: I was copying logic from the maintenance plans [here](https://github.com/rsoprivatecloud/Maintenance-Plans/blob/9461cc2b1cf59891b40a9e7128d72822109ae608/RPC%2011.x:%20Upgrade%20to%20RPC%2014.x#L280).

Frank-ZhangXin: Not sure about the correctness of the file name, since my systems don't have such file. In this case this check would fail no matter what.

BjoernT: The name only exist pre upgrade so we need to be careful as this role is version independent. The user_variables as a pre Mitaka thing so yes it will be there for older releases

rexredinger: Thanks for the heads-up, @BjoernT .  Does this maintenance step only apply to versions where user_variables.yml is still around?  Then I could just set a guard clause to skip this play when user_variables are not present.

BjoernT: Yes the maintenance plan is always specific to a version. Skipping would be non optimal too then we can better leave it inside the maintenance plan. Or we look for a different file, rpc-maas is also using

rexredinger: Looking at #138, Is that kind of change something that would make sense here?  If so, that feels like something we may want to abstract out into something like a defaults.  It feels a bit smelly to go sprinkling version knowledge across various tasks, when such knowledge applies to the entire upgrade process.  Ultimately, I want to follow my directive of automating upgrades, but avoid just creating new cruft.

rexredinger:

playbooks/roles/rpc_pre_upgrade/tasks/enable_maas_holland.yml
```diff
@@ -0,0 +1,35 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+- name: Ensure Maas is enabled
+  vars:
+    config_files:
```
rexredinger: With a better name, this could go into defaults.

playbooks/roles/rpc_pre_upgrade/tasks/enable_maas_holland.yml
```diff
@@ -0,0 +1,35 @@
...
+      - /etc/openstack_deploy/user_osa_variables_overrides.yml
+      - /etc/openstack_deploy/user_variables.yml
+  tasks:
+    - name: maas_holland_enabled is true
+      lineinfile:
+        dest: "{{ item }}"
+        regexp: "^maas_holland_enabled:"
+        line: "maas_holland_enabled: True"
+        state: present
+      with_first_found: "{{ config_files}}"
```
Frank-ZhangXin: nit, but it's ok. {{ config_files }}

PR #139, opened 2018-02-20
RLM-1437 Force kill logstash if graceful fails

antonym:
Force kills logstash if a graceful shutdown fails.  Gate jobs
were occasionally failing on trying to shutdown the service
so this attempts to shutdown gracefully before killing the
service.

### DIFF
--- a/playbooks/elasticsearch-reindex.yml
+++ b/playbooks/elasticsearch-reindex.yml
@@ -19,15 +19,18 @@
   roles:
     - role: "pip_install"
 
-- name: Stop logstash serivce
+- name: Stop logstash service
   hosts: logstash_all
   user: root
   tasks:
-    - name: Service stop
+    - name: Attempt to gracefully stop logstash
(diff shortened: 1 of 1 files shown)


playbooks/elasticsearch-reindex.yml
```diff
@@ -23,11 +23,18 @@
...
       service:
         name: logstash
         state: stopped
         pattern: logstash
+      ignore_errors: true
+    - name: Ensure logstash is stopped
+      shell: "ps -ef | grep -v grep | grep -w logstash | awk '{print $2}'"
+      register: running_processes
+    - name: Force kill logstash if still running
+      shell: "kill -9 {{ item }}"
```
BjoernT: Why not just do a pkill -f -9 logstash ?

antonym: That works too, updated to reduce the code.

PR #140, opened 2018-02-20
RLM-1438 Ensure we're using openstack_hosts fork

antonym:
Injects the openstack_hosts fork on 14.7.0 which
includes the domain fixes for RLM-322.

### DIFF
--- a/scripts/pre_redeploy.sh
+++ b/scripts/pre_redeploy.sh
@@ -135,3 +135,10 @@ cinder_api_serial: '100%'
 glance_api_serial: '100%'
 glance_registry_rolling: '100%'
 EOF
+
+# RLM-1438 patch openstack_hosts fork into 14.7.0 to ensure RLM-322 works
+if [[ ${RPC_TARGET_CHECKOUT} == "r14.7.0" ]]; then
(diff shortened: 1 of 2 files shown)


antonym:

antonym:

scripts/pre_redeploy.sh
```diff
@@ -124,3 +124,10 @@ pushd /etc/openstack_deploy/
     sed -i '/^haproxy_package_state.*/d' user_osa_variables_defaults.yml
     echo "haproxy_package_state: 'latest'" >> user_osa_variables_defaults.yml
 popd
+
+# RLM-1438 patch openstack_hosts fork into 14.7.0 to ensure RLM-322 works
+if [[ ${RPC_TARGET_CHECKOUT} == "r14.7.0" ]]; then
```
Frank-ZhangXin: Do we care about any version lower than 14.7.0?

antonym: Don't think we're deploying anything older via upgrades and this should be taken care of in 14.8.0.

claco:
Thanks @BjoernT !

PR #141, opened 2018-02-20
[RLM-1422] Move glance cache and preflight checks to pre-upgrade role

rexredinger:
I wasn't 100% sure on the continued use of marker files as is.  Let me know if there is an alternate preferred way.  Also interested if any tags could improve this.


playbooks/roles/rpc_pre_upgrade/tasks/glance-cache-cleanup.yml
```diff
@@ -13,13 +13,26 @@
...
 
-- name: Stop glance services
+- name: Stop Glance services
   hosts: glance_all
   user: root
   tasks:
-    - name: Wipe glance cache directory
+    - name: Check Glance cache has been wiped
+      stat:
+        path: "{{ marker_dir }}/glance-cach-cleanup.complete"
```
antonym: spelling on cache

playbooks/roles/rpc_pre_upgrade/tasks/main.yml
```diff
@@ -16,6 +16,30 @@
 # set backup_dir as a fact to ensure dir name doesn't change during run
 - set_fact:
     backup_dir: "{{ ansible_env.BKUPDIR |default(local_home+'/'+'rpc-upgrade-'+date_stamp) }}"
+    marker_dir: "{{ ansible_env.UPGRADE_LEAP_MARKER_FOLDER | default('/etc/openstack_deploy/upgrade-leap) }}"
```
antonym: We might need to create the /etc/openstack_deploy/upgrade-leap directory since it's likely it doesn't exist yet.

playbooks/roles/rpc_pre_upgrade/tasks/main.yml
```diff
@@ -16,6 +16,30 @@
...
+
+# Prepare Glance cache
+- include: glance-cache-cleanup.yml
+
+# Preflight checks - it'd be nice to move the idempotency logic
+# into the preflight-check.yml file itself, by maybe wrapping all
+# tasks in a conditional block.  But this is expected to run in
+# Ansible 1.9, which doesn't support that.
+- stat:
+  path: "{{ upgrade_dir}}/rpc-preflight-check.complete"
```
antonym: {{ marker_dir }} ?

antonym:
Needs a rebase on latest changes in master.

playbooks/roles/rpc_pre_upgrade/tasks/main.yml
Frank-ZhangXin: OSA ops did create the directory but I'm not sure the timing is soon enough for this role
https://github.com/openstack/openstack-ansible-ops/blob/052bdbf672dc72a9513c9103f80f76fcb7503ada/leap-upgrades/lib/functions.sh#L76

playbooks/roles/rpc_pre_upgrade/tasks/main.yml
```diff
@@ -16,6 +16,36 @@
...
+
+# Prepare Glance cache
+- include: glance-cache-cleanup.yml
+
+# Preflight checks - it'd be nice to move the idempotency logic
+# into the preflight-check.yml file itself, by maybe wrapping all
+# tasks in a conditional block.  But this is expected to run in
+# Ansible 1.9, which doesn't support that.
+- stat:
+  path: "{{ marker_dir }}/rpc-preflight-check.complete"
```
Frank-ZhangXin: path indentation reference: http://docs.ansible.com/ansible/latest/stat_module.html#examples

playbooks/roles/rpc_pre_upgrade/tasks/main.yml
Frank-ZhangXin: Miss right single quote around default path.
`default('/etc/openstack_deploy/upgrade-leap')`

playbooks/roles/rpc_pre_upgrade/tasks/main.yml
rexredinger: Good catch!

playbooks/roles/rpc_pre_upgrade/tasks/main.yml
```diff
@@ -16,6 +16,36 @@
 # set backup_dir as a fact to ensure dir name doesn't change during run
 - set_fact:
     backup_dir: "{{ ansible_env.BKUPDIR |default(local_home+'/'+'rpc-upgrade-'+date_stamp) }}"
+    marker_dir: "{{ ansible_env.UPGRADE_LEAP_MARKER_FOLDER | default('/etc/openstack_deploy/upgrade-leap') }}"
```
Frank-ZhangXin: if we set marker dir here, other tasks cannot use the fact ('marker_dir') when they are executed individually. In this case, only main task has the scope of this static variable. I suggest moving this one to rpc_pre_upgrade/defaults/main.yml. And currently glance-cache-wipe task cannot be run separately, or there would be one undefined variable.

playbooks/roles/rpc_pre_upgrade/tasks/main.yml
```diff
@@ -16,6 +16,36 @@
...
+
+- file:
```
Frank-ZhangXin: Making dir is same as above. Only main task is executed or the directory not existed.

Frank-ZhangXin:
@rexredinger rebase is needed

Frank-ZhangXin:
Is this still blocked or waiting for other PR merging?

rexredinger:
@Frank-ZhangXin Still waiting on other merges.

PR #143, opened 2018-02-20
Revert RLM-1401 and correct osa overrides use

BjoernT:
RLM-1401 introduced a unwanted sideeffect that libvirt, qemu and
other packages did not get updated on the hosts, rendering nova-
compute broken.

Also the references inside pre_redploy using
user_osa_variables_defaults.yml are changed to use the
user_osa_variables_overrides.yml as the defaults file is managed by
OSA


antonym:

PR #144, opened 2018-02-21
Adding check for pre leap stale configs

BjoernT:
A check was added to remove existing
user_deleteme_post_upgrade_variables.yml configuration prior to
running the upgrade/redeploy stage


antonym:

PR #146, opened 2018-02-27
[RLM-625] WIP: Added QC plays/roles/tasks for pre/post upgrade QC

jacobwagner:
In order to automate the pre/post upgrade QC process we needed
to identify the different QC mechanisms across the different openstack
service levels. This commit focuses on 4; environment, infrastructure,
openstack, and swift. The individual tasks check the status of the services
and output them to a file for each for a review. Each level can be called
individually as needed for whatever step (pre/post) that the user desires


jacobwagner:
This is the WIP version of what I am working on. This has basic structure

This is also the liberty branch, some things are different across which version of RPC you are on, like swift recon and venvs

I reverted back to about 2 weeks ago.

Things that I will add shortly that are being tested right now

- Error output at the end of the play and into an error log
- Actual Disk space checking on all hosts
- VM creation as per these steps https://github.com/rsoprivatecloud/Maintenance-Plans/blob/master/RPC%2011.x:%20Upgrade%20to%20RPC%2014.x#L591-L610
- Tempest Runs
- Rally Runs

I should have all this tested and this PR updated by EOD tomorrow

claco:
@BjoernT @Komish @jordan Work in Progress. Can you all start to double check this when you have time to make sure we don't do any unsafe things in customer environments?

Thanks!

jacobwagner:
Here is the error logging work so far

https://github.com/jacobwagner/rpc-upgrades/tree/qc-error-logs

should be done soon

PR #147, opened 2018-03-01
Fix syntax and logic issues around enable_maas_holland.yml

BjoernT:
PR 135 passed gating but included several syntax issues which are addressed with this fix


rexredinger:
Thanks for the changes @BjoernT !  If config_files is being set as a fact and possibly used elsewhere, would it be more useful to move it to defaults and rename it to something more descriptive?  Something like openstack_deploy_variables_files.

BjoernT:
No this is just a temporary fact we don't need it set as default yet

PR #151, opened 2018-04-10
RLM-268 Set group_vars for cinder CLI

Frank-ZhangXin:
Existing cinder CLI command in newton cinder role has wrong logic to
acquire '--insecure' flag under self signed cert SSL connection.
Manually pass 'true' value to make logic working.
Jira page of this bug: https://rpc-openstack.atlassian.net/browse/RLM-268

### DIFF
--- a/scripts/pre_redeploy.sh
+++ b/scripts/pre_redeploy.sh
@@ -142,3 +142,10 @@ if [[ ${RPC_TARGET_CHECKOUT} == "r14.7.0" ]]; then
      patch -p1 < ${RPC_UPGRADES_DEFAULT_FOLDER}/playbooks/patches/newton/rpc-o-14.7.0-ansible-role-requirements.patch
    popd
 fi
+
+#RLM-268 Manually give true value to the logic so cinder CLI can have --insecure flag
+mkdir -p /etc/openstack_deploy/group_vars
+if [[ -f "/etc/openstack_deploy/group_vars/cinder_all.yml" ]]; then
+    sed -i '/^keystone_service_adminuri_insecure.*/d' /etc/openstack_deploy/group_vars/cinder_all.yml
+fi
+echo 'keystone_service_adminuri_insecure: true' | tee -a /etc/openstack_deploy/group_vars/cinder_all.yml


scripts/pre_redeploy.sh
```diff
@@ -142,3 +142,7 @@ if [[ ${RPC_TARGET_CHECKOUT} == "r14.7.0" ]]; then
      patch -p1 < ${RPC_UPGRADES_DEFAULT_FOLDER}/playbooks/patches/newton/rpc-o-14.7.0-ansible-role-requirements.patch
    popd
 fi
+
+LM-268
```
rexredinger: Missing "#R"

scripts/pre_redeploy.sh
```diff
@@ -142,3 +142,7 @@ if [[ ${RPC_TARGET_CHECKOUT} == "r14.7.0" ]]; then
      patch -p1 < ${RPC_UPGRADES_DEFAULT_FOLDER}/playbooks/patches/newton/rpc-o-14.7.0-ansible-role-requirements.patch
    popd
 fi
+
+#RLM-268 Manually give true value to the logic so cinder CLI can have --insecure flag
+mkdir -p /etc/openstack_deploy/group_vars
+echo 'keystone_service_adminuri_insecure: true' | tee -a /etc/openstack_deploy/group_vars/cinder_all.yml
```
mendeni: Should this be idempotent?  As written, this will (try to) create `group_vars` dir and append to `cinder_all.yml` each time it's invoked.

Frank-ZhangXin: Yeah, you're right. 'mkdir -p' can keep idempotent but 'echo' needs a pre check for the key-value pair.

BjoernT:
@Frank-ZhangXin Wasn't the plan to fork the role so we don't have to misuse keystone_service_adminuri_insecure ? Last I tested, enabling  keystone_service_adminuri_insecure caused other issues with the API, especially admin endpoints are usually never SSL at this point

Frank-ZhangXin:

Frank-ZhangXin:

BjoernT:
Closing as the issue got closed

PR #152, opened 2018-04-12
Fix r14.7.0 role patch idempotency

BjoernT:
The live patching of the r14.7.0 ansible role broke idempotency.
A check is added to skip patching when successful

Closes-Bug: RLM-1451

### DIFF
--- a/scripts/pre_redeploy.sh
+++ b/scripts/pre_redeploy.sh
@@ -32,13 +32,13 @@ export OA_OVERRIDES="${OA_OVERRIDES:-/etc/openstack_deploy/user_osa_variables_ov
 warning "Please DO NOT interrupt this process."
 notice "Pre redeploy steps"
 pushd ${LEAPFROG_DIR}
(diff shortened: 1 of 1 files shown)


scripts/pre_redeploy.sh
```diff
@@ -137,8 +137,9 @@ glance_registry_rolling: '100%'
 EOF
 
 # RLM-1438 patch openstack_hosts fork into 14.7.0 to ensure RLM-322 works
-if [[ ${RPC_TARGET_CHECKOUT} == "r14.7.0" ]]; then
+if [ ${RPC_TARGET_CHECKOUT} = "r14.7.0" -a ! -f "${UPGRADE_LEAP_MARKER_FOLDER}/patch-rpc-o-14.7.0-ansible-role-requirements.complete" ]; then
```
rexredinger: I could be wrong, but I think `${RPC_TARGET_CHECKOUT}` needs to be quoted if you are using single  bracket test notation.

BjoernT: Nope but it's best practice, let me add it. It's important once the value is not predefined

PR #156, opened 2018-04-30
RLM-1456 Force user to export release version

Frank-ZhangXin:
Ideally users don't need worry about release upgrade from since
openstack-release can give the version. In fact at most of time it
cannot guarantee giving proper release info, especially the file not on
infra node or outdated. This check will force user to export correct
release name.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -68,6 +68,12 @@ the upgrade by setting these environment variables:
 
    export RPC_TARGET_CHECKOUT=r14.7.0
 
+If you cannot locate `/etc/openstack-release` or it is outdated. Export the release version which upgrade from manually:
+
+.. code-block:: shell
+
+    export CODE_UPGRADE_FROM='KILO/LIBERTY'
+
 The next step is to execute the leapfrog upgrade script and follow the prompts:
 
 .. code-block:: shell
--- a/playbooks/preflight-check.yml
+++ b/playbooks/preflight-check.yml
@@ -21,19 +21,26 @@
     - name: Ping check
       ping:
 
-- name: Openstack release check
+- name: Openstack release check if exists 
   hosts: localhost
   gather_facts: false
   user: root
   tasks:
-    - name: Get Openstack release
-      shell: grep 'DISTRIB_RELEASE' /etc/openstack-release | sed -e 's/.*"\(.*\)"/\1/'
-      register: openstack_release_return
+    - stat:
+        path: /etc/openstack-release
+      register: openstack_release_stat
 
-    - name: Check Openstack release
+    - stat:
+        path: /etc/rpc-release
+      register: rpc_release_stat
+
+    - name: Check release upgrade from
       fail:
-        msg: "Current Openstack release version is higher than Newton, and leapfrog isn't appliable."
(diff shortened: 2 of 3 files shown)


rexredinger:
My understanding of the problem is that the first pass at running the upgrade script sets `CODE_UPGRADE_FROM='KILO'`.  Then, the upgrade script fails after setting up Liberty, but `CODE_UPGRADE_FROM` is still set to KILO.  Rerunning the upgrade script, it sees the old `CODE_UPGRADE_FROM` environment variable (set to KILO) and tries to do a bunch of setup that is already done.

This PR is good because it helps us bypass the unreliable `discover_code_version` function on the first pass, but it does not solve the problem in the ticket (where rerunning upgrades breaks).

I think this solution needs to do one of these things:

- the upgrade scripts need to set `CODE_UPGRADE_FROM` to the new OS version before reaching the redeploy phase

- Or we need to zero out 'CODE_UPGRADE_FROM` at the start of every upgrade script execution (before the ansible check in this PR).

BjoernT:
I think the issue is that we have in the end upgrade markers and code_upgrade_from tracking similar things and now become out of sync. Ideally, code_upgrade_from should be fixed, in the same way code_upgrade_to is and the markers would do the primary job of tracking progress

Frank-ZhangXin:
Copied comment from JIRA card.

Talked with engineer folks about release missing issue of this one. The conclusion is 'openstack_release' file is a must-have one and it has to be correct before starting leapfrog upgrade, since it has been used multiple places and very complex to change all of them to adapt a new way to determine which release it it now.

Current leapfrog will fail to give right release tag at db upgrade part if db upgrade failed. '$CODE_UPGRADE_FROM' would fall back to base release before upgrade begins. Since '$CODE_UPGRADE_FROM' discovering depends on openstack_release file which won't change until redeploy starting. The work around (I really can't name it 'solution'), add task to define right value to '$CODE_UPGRADE_FROM' by searching 'db-migration-<mitaka/newton>-ymlxxx' marker file. While it can only deal with this card particular case which failed at newton DB migration.

So there's still possibility that leapfrog mis-set '$CODE_UPGRADE_FROM' before redeploy's opesntack_host role done.

antonym:

Frank-ZhangXin:
Note: neutron-remove-old-containers.yml failed

Frank-ZhangXin:
kilo gate:
18:11:10 fatal: [prumtasktnl-444-b91e]: FAILED! => {"changed": true, "cmd": ["chroot", "/var/lib/lxc/LXC_NAME/rootfs", "/usr/local/bin/cache-prep-commands.sh"], "delta": "0:00:00.005752", "end": "2018-05-02 23:11:10.416530", "failed": true, "rc": 1, "start": "2018-05-02 23:11:10.410778", "stderr": "+ '[' -a /etc/resolv.conf ']'\n+ echo 'nameserver 10.255.255.1'\n/usr/local/bin/cache-prep-commands.sh: line 8: /etc/resolv.conf: No such file or directory", "stdout": "", "stdout_lines": [], "warnings": []}

antonym:
recheck_all

PR #158, opened 2018-05-03
RLM-1374  Malformed YAML 2: Electric Boogaloo

rexredinger:
This ensures that the user config starts with `---`

The added ` explicit_end=False,` is added in hopes that it might address sightings of `---` in places other than the beginning of the document.


antonym:
recheck_all

PR #161, opened 2018-05-04
[RLM-1460] Add check for override config ops

cloudnull:
If any override options existin in config we should halt the playbook
run and have the deployer evaluate the config before going forward. If
the operator has determined that the playbook run will be successful
with the override options in place they can set
"leap_upgrade_advanced_configuration_check_skip" to true.

Issue: RLM-1460
Signed-off-by: Kevin Carter <kevin.carter@rackspace.com>

### DIFF
--- a/playbooks/preflight-check.yml
+++ b/playbooks/preflight-check.yml
@@ -192,8 +192,36 @@
         msg: >-
           CHECK FAILED: Glance currently has an NFS mount but has no user
           configuration to support it. Please review the documentation on
-          setting up glance with NFS - 
+          setting up glance with NFS -
           https://docs.openstack.org/openstack-ansible-os_glance/latest
-      when: 
+      when:
         - nfs_mount_found.rc == 0
         - glance_nfs_client is undefined
+
+- name: Check for any advanced configuration options
+  hosts: localhost
+  connection: local
+  gather_facts: false
+  user: root
+  tasks:
+    - name: Check for known advanced configuration options
+      shell: >-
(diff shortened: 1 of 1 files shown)


antonym:
Gate is failing with:
```
14:59:33 TASK: [Fail if advanced configuration has been found] ************************* 
14:59:33 failed: [localhost] => {"failed": true}
14:59:33 msg: Advanced configuration options have been defined within the user variables found on this system. Please review the following options before continuing, "file_path:line_number:option" - [u'/etc/openstack_deploy/user_extras_variables.yml:49:tempest_tempest_conf_overrides', u'/etc/openstack_deploy/user_variables.yml:76:neutron_l3_agent_ini_overrides', u'/etc/openstack_deploy/user_variables.yml:80:neutron_ml2_conf_ini_overrides', u'/etc/openstack_deploy/user_variables.yml:83:neutron_neutron_conf_overrides', u'/etc/openstack_deploy/user_variables.yml:162:repo_build_upper_constraints_overrides']. If these options have been audited and known to be compatible with the targeted upgrade release set the extra variable "leap_upgrade_advanced_configuration_check_skip" to true in any user variable file or on the command line using "-e 'leap_upgrade_advanced_configuration_check_skip=true'".
```

Is this something that needs to be skipped during gate runs as I think the AIO/leapfrog has some overrides by default?

BjoernT:
Hmm, I don't necessarily want to abort every upgrade "by default" due to config_template usage, which is given quite often. I think we are better off on addressing known config issues directly, as part of the product and fixing the facts insertion via https://review.openstack.org/#/c/566360/ rather than blocking the upgrade

PR #163, opened 2018-05-09
RLM-1470 Update gating post to leverage RPC-O scripts

antonym:
RPC-O gating scripts were updated to gather logs from
MNAIO and AIO builds.  This brings those scripts to
rpc-upgrades.  The previous iteration was pulled from
RPC-O as well.

### DIFF
--- a/gating/pre_merge_test/collect_logs.yml
+++ b/gating/pre_merge_test/collect_logs.yml
@@ -1,0 +1,147 @@
+- hosts: "{{ target_hosts }}"
+  gather_facts: false
+  tasks:
+    - name: Ensure artefacts directories exists
+      file:
+        path: "{{ item }}"
+        state: directory
+      with_items:
+        - "{{ artifacts_dir }}/{{ inventory_hostname }}/host"
+        - "{{ artifacts_dir }}/{{ inventory_hostname }}/containers"
+      delegate_to: "localhost"
+
+    - name: Grab host data
+      command: >
+               rsync
+               --archive
+               --compress
+               --verbose
+               --relative
+               --rsh 'ssh -o StrictHostKeyChecking=no'
+               --ignore-missing-args
+               --safe-links
+               --no-perms
+               --no-owner
+               --no-group
+               {{ inventory_hostname }}:{{ item }}
+               {{ artifacts_dir }}/{{ inventory_hostname }}/host
+      with_items:
+        - "/openstack/log"
+        - "/etc"
+        - "/var/./log"
+      delegate_to: "localhost"
+      ignore_errors: true
+      tags:
+        - skip_ansible_lint
+
+    - name: List containers
+      command: "lxc-ls -1"
+      failed_when:
+        - containers.rc != 0
+        - containers.msg != '[Errno 2] No such file or directory'
+      changed_when: false
+      register: containers
+
+    - name: Get container PIDs
+      command: "lxc-info --name {{ item }} --no-humanize --pid"
+      with_items:
+        - "{{ containers.stdout_lines | default([]) }}"
+      register: container_pids
+
+    - name: Grab container data
+      command: >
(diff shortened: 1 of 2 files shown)


gating/pre_merge_test/collect_logs.yml
```diff
@@ -0,0 +1,147 @@
...
+               --no-owner
+               --no-group
+               {{ inventory_hostname }}:/proc/{{ item[0].stdout }}/root/{{ item[1] }}
+               {{ artifacts_dir }}/{{ inventory_hostname }}/containers/{{ item[0].item }}
+      when:
+        - containers.rc == 0
+        - item[0].stdout != ""
+      with_nested:
+        - "{{ container_pids.results }}"
+        -
```
Frank-ZhangXin: Can you explain why it uses sub layer list here?

antonym:
Looks like the log artifacts got generated properly and uploaded.  Think this is good to proceed with:

https://gist.github.com/antonym/5abc72114d23649d905e6a6c178a7324

gating/pre_merge_test/collect_logs.yml
```diff
@@ -0,0 +1,147 @@
...
+      file:
+        path: "{{ item }}"
+        state: directory
+      with_items:
+        - "{{ artifacts_dir }}/{{ inventory_hostname }}/host"
+        - "{{ artifacts_dir }}/{{ inventory_hostname }}/containers"
+      delegate_to: "localhost"
+
+    - name: Grab host data
+      command: >
```
BjoernT: Is there are reason for not using the synchronize module ?

gating/pre_merge_test/collect_logs.yml
```diff
@@ -0,0 +1,147 @@
...
+      register: containers
+
+    - name: Get container PIDs
+      command: "lxc-info --name {{ item }} --no-humanize --pid"
+      with_items:
+        - "{{ containers.stdout_lines | default([]) }}"
+      register: container_pids
+
+    - name: Grab container data
+      command: >
```
BjoernT: Same as above

gating/pre_merge_test/collect_logs.yml
```diff
@@ -0,0 +1,147 @@
...
+  gather_facts: false
+  tasks:
+    - name: Find kibana-selenium screenshots
+      find:
+        paths: "/opt/kibana-selenium"
+        patterns: "*.png"
+      when: inventory_hostname in ["infra1", "localhost"]
+      register: kibana_selenium_screenshots
+
+    - name: Ensure kibana artefacts directory exists
```
BjoernT: update to artifact

gating/pre_merge_test/collect_logs.yml
rexredinger: I think what @Frank-ZhangXin  is asking is why `"etc"` and `"var/log"` are in a list of their own, rather than being first tier elements of the list.  I think the answer is that `with_nested` expects a list of lists.  More info [here](https://github.com/ansible/ansible/blob/devel/lib/ansible/plugins/lookup/nested.py).

gating/pre_merge_test/collect_logs.yml
antonym: Not really, I copied this exactly from:

https://github.com/rcbops/rpc-openstack/blob/newton/gating/pre_merge_test/collect_logs.yml

So if we want to make changes, I'd recommend we make them there and port the changes here.  They do the same actions so I'd like to keep them aligned if it all possible.

gating/pre_merge_test/collect_logs.yml
Frank-ZhangXin: yeah, document of nest did make sense to me. 'Compose list of lists.' I have no question about this line now.

PR #172, opened 2018-05-31
ASC-564 Persist release envvars for test results

johnduarte:
In order to capture accurate metadata about Molecule system tests, the
"RPC_RELEASE" and "RPC_PRODUCT_RELEASE" environment variables need to be
stored in the "MNAIO_VAR_FILE". Values in this file are re-sourced by the
`run` script to ensure they are used by the system tests in reporting
their results.


odyssey4me:
FYI "RPC_RELEASE" and "RPC_PRODUCT_RELEASE" are not consistently available across all releases. The second is pike+ and the first I think is available for newton, not sure about older.

johnduarte:
@ryan-rs see Jesse's comment above. We may need to work around this.

antonym:

PR #174, opened 2018-06-01
Workaround for deployment host != infra host

BjoernT:
Adding a workaround when deployment host is not the same as a infra host
and the resulting issues around lookup/sync of the `/etc/openstack-release`
file


Frank-ZhangXin:

PR #176, opened 2018-06-05
RLM-1471 Adds playbook for prepping incrementals

antonym:
Adds playbook for prepping incremental upgrades, fixes up
repos, brings all packages to latest, fixes up configs before
running upgrades.

### DIFF
--- a/playbooks/prepare-ocata-upgrade.j2
+++ b/playbooks/prepare-ocata-upgrade.j2
@@ -1,0 +1,12 @@
+# {{ ansible_managed }}
+
+# Base repository
+deb http://rackspace.clouds.archive.ubuntu.com/ubuntu/ xenial main universe
+
+# Security Updates
+deb http://security.ubuntu.com/ubuntu xenial-security main universe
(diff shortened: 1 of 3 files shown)


antonym:
Merging to test in Jenkins

BjoernT:
This likely will cause problems where customers force us to use their own apt mirrors.
Replacing the sources.list  is a problem here, why was this chosen ?
Or is this play just for testing ?

antonym:
@BjoernT Yeah, I hardcoded it for testing, ideally I'll pull out the sources, but the issue I was running into was that I was getting dependency issues installing python3-dev during Ocata because the artifact repos and the basic repos were set preventing me from doing a full proper upgrade of all packages.

We currently install this one during newton which is missing some needed update repos:

https://github.com/rcbops/rpc-openstack/blob/newton/rpcd/playbooks/configure-apt-sources.j2

If that is in place, I can't upgrade packages to the latest which causes that mismatch to occur.  We probably need to add the additional sources to the apt repo above.

PR #177, opened 2018-06-05
RLM-1496 Update gating hooks to reflect new style

antonym:
pre_merge_test becomes check, post_merge_test becomes periodic


mattt416:
Thanks @antonym !

antonym:
recheck_all

antonym:
recheck_all

PR #185, opened 2018-06-11
Move apt update earlier before bootstrap of Ocata

antonym:
This allows containers to be updated without hitting
the container inventory that hasn't been provisioned
yet in Ocata.


antonym:
Merging to gate test.

PR #187, opened 2018-06-13
Fix non working check for CODE_UPGRADE_FROM

antonym:
Existing check was not picking up correctly and not setting
variables correctly.


antonym:

antonym:
recheck_all

PR #188, opened 2018-06-15
FLEEK-61 Fixes for Mitaka Leapfrog

antonym:
FLEEK-61 Adds Mitaka to Newton Support in rpc-upgrades

This commit enables support for running a Mitaka
upgrade and contains support needed for running
Mitaka AIO and MNAIO gating jobs.  It contains the
following fixes:

- Skips the variable migration for Mitaka
- Removes user_osa_variables_defaults as existing
settings don't work in Newton and adds neccessary
variables back into file.
- Disables Ceilometer in AIO, swift breaks because
it depends on Ceilometer running.
- Sets haproxy fix to run on AIOs only, not MNAIOs
Looks like Mitaka gate was hitting this as well, so
this ports the patch over to Mitaka.
- Pushes updated SPICE url to MNAIO infra1
- Ensures openstack_release is set to rpc_release to
avoid venv_tags from getting set to newton-eol instead
of the latest RPC release version.
- Adds release note

Depends on https://review.openstack.org/#/c/575527/ which has merged.

### DIFF
--- a/playbooks/patches/mitaka/haproxy-symlink-fix.patch
+++ b/playbooks/patches/mitaka/haproxy-symlink-fix.patch
@@ -1,0 +1,38 @@
+diff --git a/playbooks/roles/haproxy_server/tasks/haproxy_pre_install.yml b/playbooks/roles/haproxy_server/tasks/haproxy_pre_install.yml
+index adea9278..4f4b50eb 100644
+--- a/playbooks/roles/haproxy_server/tasks/haproxy_pre_install.yml
++++ b/playbooks/roles/haproxy_server/tasks/haproxy_pre_install.yml
+@@ -13,29 +13,13 @@
+ # See the License for the specific language governing permissions and
+ # limitations under the License.
+ 
+-- name: Test for log directory or link
+-  shell: |
+-    if [ -h "/var/log/haproxy"  ]; then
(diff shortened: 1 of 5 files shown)


antonym:

antonym:

antonym:
Mainly that it uses the leapfrog code instead of the incremental major upgrade code.  The leapfrog tears down the containers and does the upgrade while everything is off-line.  It's a cleaner way to do the upgrade as it starts with everything fresh besides the data/instances/storage.  It was preferred since there's a lot of cruft from the older versions and we wanted to clean things up for support and ops.

It's probably worth us dropping the Mitaka to Newton major upgrade code (ubuntu14-mitaka-to-newton.sh) since we're going the leapfrog route for this version.

rexredinger:
Makes sense to me.  To review, a leapfrog:

1. Uses rpc-upgrades leapfrog code, and does not use provided OSA upgrade scripts.
2. Completely tears down and rebuilds the containers (ie. involves downtime)

and we are going the route to avoid likely crufty and unstable scenarios.

antonym:
Yeah, it's a mix of the code in rpc-upgrades and the leapfrog code up at https://github.com/openstack/openstack-ansible-ops/tree/master/leap-upgrades

PR #190, opened 2018-06-15
RO-4265 Simplify spice url patch

antonym:
Uses a sed instead of a patch to avoid variations
across different versions.


antonym:
Checks passing, merging in.

PR #194, opened 2018-06-27
FLEEK-84 Various fixes for deploying older Mitaka

antonym:
FLEEK-84 Various fixes for deploying older Mitaka

* Patch in horizon install branch in Mitaka
* Uses mitaka-eol galera_server which contains fixes for install
* Ensures galera/rabbit are only deployed with a single affinity


antonym:
Merging changes to test in Mitaka gates.

PR #197, opened 2018-06-29
FLEEK-125 Use older setuptools for Mitaka installs

antonym:
Uses older setuptools if bad version found


antonym:
Merging to test in gates.

PR #201, opened 2018-07-03
Testing downgrade setuptools to 33.1.1

antonym:
Aligns with global pins in newton-eol


antonym:
Testing on my branches

PR #202, opened 2018-07-03
Reset default apt sources after installing libvirt-dev

antonym:
Moves reset default apt sources to after installing libvirt-dev.  Fixes kilo issue on MNAIO.


antonym:
Fixes MNAIO Kilo job.

PR #207, opened 2018-07-13
FLEEK-142 Use rcbops galera_server branch for Mitaka

antonym:
Uses rcbops galera_server branch as we needed to bump
percona-xtrabackup version for testing the initial Mitaka
deploy.  Without this version bump, Galera hangs during
cluster installation.


antonym:
Merging to test in gate.

PR #208, opened 2018-07-13
[WIP] RI-347 N2Q Tempest testing

rexredinger:
This is a first pass at getting some tempest tests back into rpc-upgrades gating.

Outstanding questions:

- What is the preferred method of collecting output from Tempest tests (assuming one exists)?
- Are there any desired test scenarios beyond the defaults provided by os-tempest role?
- Do we have any interest in parallelizing the tests (should I look into that)?


rexredinger:
Actually just noticed the collect_logs.yml file under gating/check.  I'm going to add a snippet in there to grab tempest output from the utility container.

rexredinger:
It should be complete enough to merge and start trying it out. However, I have a concern that the current setup won't provide good feedback about what part of Tempest failed, when failures occur.  I can add log collection in a separate PR though.  So let me just squash these commit and I'll merge this PR.

PR #211, opened 2018-07-18
Skips the ocata setup-openstack script during N->Q

antonym:
Brings down the upstream scripts so that they can be modified
and optimized for leaping over Ocata to save some
deployment cycles


antonym:
Merging to gate test.

PR #212, opened 2018-07-19
Going back to upstream scripts for N to Q

antonym:
Using sed to rip out steps to skip


antonym:
Merging to gate test.

PR #218, opened 2018-08-03
FLEEK-151 Fix enable_lxc_fs_check flag

antonym:
Previously if the flag was set to disable it would skip the
task to gather the data, but the following task would fail because
that data was missing.  This sets it to always gather that
information, tosses an informational alert and skips the hard failure.


antonym:

PR #219, opened 2018-08-08
FLEEK-155 Ensure apt is in a healthy state during preflight

antonym:
Runs a dpkg -a --configure; apt install -y -f to ensure all
package installs across all containers and hosts are clean so
the upgrade process isn't hindered by package failures from
apt being in a half completed state.


antonym:
recheck_all

PR #221, opened 2018-08-10
FLEEK-152 Remove security hardening from process

antonym:
Has been dropped from the upgrade process, so this removes
some cruft.


odyssey4me:
recheck_container_ansible-lint_tox-test

PR #229, opened 2018-08-23
Use rackspace mirror for apt sources

odyssey4me:
MNAIO images use archive.ubuntu.com which is slow and fails
more often due to there being more hops to it. Instead we
prefer to use the Rackspace mirrors which are local to each
data centre and have turned out to be more reliable.


odyssey4me:

PR #232, opened 2018-08-30
Testing fix for queens upgrades, using local fork

antonym:
Current queens script doesn't exit properly so adding
in the fix here to test queens in gate.


antonym:
Merging to gate test.

PR #234, opened 2018-08-31
RI-454 Fall back to master maas vars if release isn't found

antonym:


antonym:
Merging to gate test.

PR #235, opened 2018-09-02
Drop user_rpco_variables_defaults.yml for N->O

antonym:
Variables there aren't valid in newer releases


antonym:
Merging to test in gates

PR #236, opened 2018-09-04
Set default branch to newton-eol

antonym:
If master if used by default, it has issues
with 14.04 since it assumes systemd is available.


antonym:
Merging since PR tests don't test MNAIO.

PR #238, opened 2018-09-10
Remove cleanup now that bootstrap is not run

antonym:
See https://review.openstack.org/#/c/600875/


antonym:
Merging to gate test.

PR #240, opened 2018-09-13
Adjust generate_reno_report to handle tag failure

antonym:
Script pulled from rpc-o which includes fixes to allow gate to pass.


antonym:
Merging, this should allow the release gates to flip green instead of marking red from release notes not passing.

PR #241, opened 2018-09-14
RI-466 Remove ansible_facts to clean old facts out

antonym:
Old facts remain after the inventory has been updated.
This removes the facts and forces a rebuild of them the
next time a run happens.


antonym:
Merging to test in N2Q gate.

PR #245, opened 2018-09-26
Set password file location

antonym:
Set password file location so upgrade process doesn't fail when it can't find the proper password file name


BjoernT:
This will not work moving forward when ansible vaults are deployed. Every code looking/modifying *secrets*.yml has to be updated to accommodate vauits

BjoernT:
Yes the user-secrets-adjustment is a problem. We have only two options, either decrypt or take those steps out of OSA and add them to the release notes only

PR #246, opened 2018-09-28
RI-495 Fix tempest tests to run on Queens upgrade

antonym:
Removes all tempest settings from various files and
centralize them into user_rpco_tempest.yml for gating
purposes.  Also whitelists basic tests to run.


antonym:
Merging to test in Queens upgrade gates.

PR #247, opened 2018-09-28
Update MNAIO disk to use

antonym:
DATA_DISK_DEVICE is no longer present so use new method


antonym:
Merging to gate test in MNAIO

PR #248, opened 2018-10-01
Ensure tempest_images is not overridden

antonym:
Older versions are missing additional settings like disk format
which causes tempest to fail when uploading images to glance.


antonym:
Merging to test in N2Q gate.

PR #260, opened 2018-10-08
Pin to specific get-pip to avoid instabilities upstream

antonym:
Pin to specific get-pip to avoid instabilities upstream


antonym:

antonym:
Fixes gates, merging.

PR #264, opened 2018-10-16
Hacking in some testing galera/keystone fixes (N2Q)

antonym:


antonym:
Merging to gate test.

PR #276, opened 2018-11-02
Adds the ability to capture images pre upgrade

antonym:
By setting a MNAIO upgrade job with the action of snapcap
it will skip the upgrade portion and capture the images of
the environment before the upgrade is ran if the deployment is
successful.  This way we can leverage the workarounds in
rpc-upgrades to deploy a snapshot state of older versions and
leverage those when testing upgrades.


antonym:
recheck_whisperer

antonym:
recheck_all

PR #278, opened 2018-11-07
Get version of job context for artifact upload

antonym:
Adjust vars for artifact upload


antonym:
Merging for gate run.

PR #281, opened 2018-11-12
Only run functions if available, run ARA log collection

antonym:
Adds a check to load functions if available, so it doesn't
break on versions that don't have that file (older)

Also adds a check to run ARA if the binary is detected.  Older
versions did not support ARA, so if the binary isn't present, it
will skip ARA collection.


antonym:
Merging to test.

PR #283, opened 2018-11-27
Uses vars from rpc-openstack for upgrades

antonym:
Syncs group_vars and runs the site-release playbook
from rpc-openstack to ensure the proper variables for
a supported release are used during an upgrade.

This ensures the proper RPC release version numbers are set
and that any overrides we've made to the release are used
during the upgrade.


antonym:

PR #284, opened 2018-11-27
Ensure we always have the latest repo

antonym:
When using snapshot builds, the head branches may
be out of date and can check out invalid code.

This refreshes the repo before upgrade to make sure
we're running the correct code.


antonym:
recheck_all

antonym:
Recheck_all

PR #285, opened 2018-11-27
RI-575 Fix sshd tag on deployment of old versions

antonym:


antonym:
Upstream fix has merged, https://github.com/openstack/openstack-ansible-ops/commit/2ec3466ad8875572a7bad42902127983fffd3165, repointing at upstream instead of local branch and merging to fix gating/leapfrogs.

PR #290, opened 2018-12-05
RI-467 Uses a case to handle deploys for diff vers

antonym:
The style of deploy changes post newton, so we adjust
the create-mnaio script to account for the variations
in deployment.


antonym:
Merging for gate testing.

PR #293, opened 2018-12-07
Set RPC_APT_ARTIFACT_MODE="loose" for jobs

antonym:
Ensures newton jobs always deploy inital env
with loose mode


antonym:

PR #302, opened 2018-12-19
[RI-624] Set distribution variable before upgrades

rexredinger:
Manages the update of rpc_relase


antonym:
This will hopefully be covered by https://github.com/rcbops/rpc-upgrades/pull/308 instead as there are additional files and management that need to be done to manage the upgraded releases.

antonym:
Closing, should be covered by https://github.com/rcbops/rpc-upgrades/pull/308

PR #303, opened 2018-12-19
[RI-621] Only prep Ocata once at most

rexredinger:
Also allows manual override by operator.
The reason for this is to avoid issues with failed reruns,
sometimes manifesting as template failures.  It also saves some
time on reruns.

### DIFF
--- a/incremental/lib/functions.sh
+++ b/incremental/lib/functions.sh
@@ -204,9 +204,11 @@ function strip_install_steps {
 }
 
 function prepare_ocata {
-  pushd /opt/rpc-upgrades/incremental/playbooks
-    openstack-ansible prepare-ocata-upgrade.yml
-  popd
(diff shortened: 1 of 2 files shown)


antonym:
In order to make this idempotent, can we touch the file at the end of the prepare-ocata-upgrade.yml playbook instead?  That way it'll at least attempt to rerun if the playbook fails.  

Let's also remove https://github.com/rcbops/rpc-upgrades/blob/master/incremental/playbooks/prepare-ocata-upgrade.yml#L41 (Update all packages to the latest version) as the upgrade process shouldn't handle package updates.

antonym:
FYI, the P2Q gate won't catch errors with this as it'll only run on the Newton build, so we'll need to test manually.

antonym:
recheck_container_bashate_tox-test

rexredinger:
quick whitespace fix.

antonym:
Cool, I tested manually and it worked, all tests passing and P2Q test won't touch it so I'll merge this.

PR #304, opened 2018-12-19
[RI-620] (WIP) Add an optional play to disable IPv6 on containers

rexredinger:
As per RI-620, we ran into an issue where containers were using ipv6 interfaces and failing to install packages as a result.  This pull aims to allow operators to disable IPv6 for containers, if they deem it necessary.


BjoernT:
The problem I see here this does not survive container recreation and what if you actually run ipv6 networks in neutron, like customers could do.

PR #312, opened 2019-01-10
FLEEK-243 Adds Ocata Cell Generation and Nova/Cinder migrations

antonym:
Adds Cells v2 DB
Adds initial Cells v2 cells
Does Cinder/Nova migration from Ocata venv
Removes deletion of openstack-ansible wrapper in Ocata prepare as
it should be overwritten during the next bootstap

### DIFF
--- a/incremental/playbooks/create-cell0.yml
+++ b/incremental/playbooks/create-cell0.yml
@@ -1,0 +1,44 @@
+---
+
+- name: Ensure nova_cell0 database is created
+  hosts: "galera[0]"
+  user: root
+  gather_facts: false
+  tasks:
+    - name: Ensure nova_cell0 is created
+      mysql_db:
+        login_user: "{{ galera_root_user }}"
+        login_password: "{{ galera_root_password }}"
+        login_host: "{{ nova_galera_address }}"
+        name: "{{ nova_cell0_database }}"
+        state: present
+
+    - name: Grant access to DBs for the services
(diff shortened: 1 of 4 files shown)


incremental/playbooks/create-cell0.yml
```diff
@@ -0,0 +1,44 @@
...
+        login_user: "{{ galera_root_user }}"
+        login_password: "{{ galera_root_password }}"
+        login_host: "{{ nova_galera_address }}"
+        name: "{{ item['name'] }}"
+        password: "{{ item['password'] }}"
+        host: "{{ item['host'] }}"
+        state: "present"
+        priv: "{{ item['database'] }}.*:ALL"
+        append_privs: "{{ item['db_append_privs'] | default(omit) }}"
+      with_items:
```
rexredinger: Could the with_items clause be simplified to only list the host entries?

antonym: Yeah, it probably could.  I copied some of this from upstream and it was set to do additional items as well so there's probably a few things we could compress.  It worked though so I haven't tried optimizing it yet.

antonym:

antonym:

antonym:
Merging this in... I've added N2P tests in addtion to P2Q that will now gate on PRs so we'll get some more exercising on this with a few more PRs I'm putting up.

PR #313, opened 2019-01-14
WIP Leap v2 POC (Newton and On)

antonym:
Testing, up for comments

### DIFF
--- a/incremental/leapfrog-upgrade.sh
+++ b/incremental/leapfrog-upgrade.sh
@@ -1,0 +1,151 @@
+#!/usr/bin/env bash
+
+# Copyright 2018, Rackspace US, Inc.
+#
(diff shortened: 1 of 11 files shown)


BjoernT:
@antonym What is the reason for shutting down the database ? Also neutron-server is stopped as precaution or known incompatibility of DB vs current running service

antonym:
@BjoernT So during the upgrade from 10.0 to 10.1 the playbooks were getting confused about who had quorum and would bomb during the upgrade.  We found by shutting all nodes except the primary, we could ensure the first node had quorum and it seemed to help the 10.1 upgrade complete.

https://review.openstack.org/#/c/611438/

antonym:
Closing this, focusing on https://github.com/antonym/osa-warp to handle this instead.

PR #314, opened 2019-01-14
Queens repo containers cleanup

BjoernT:
As preparation for the queens upgrade, the
repo containers are be destroyed in order to free up
disk space occupied by previous upgrades.
This reduces potential issues, where the root device runs
out of disk space


antonym:
I'll try another attempt at this but I'm guessing because the pip-conf-removal.yml attempts to gather facts before setup-host occurs, it gets mad because the repo containers are missing.  Because it's just removing a file we could probably get away without gathering facts so that may be an upstream change to set gather_facts to false or an ignore_errors we make so that we can then can freely remove containers ahead of the upgrade if we want to refresh certain ones.

https://github.com/openstack/openstack-ansible/blob/stable/queens/scripts/upgrade-utilities/playbooks/pip-conf-removal.yml#L18

PR #316, opened 2019-01-21
Exclude CEPH nodes from certain pre upgrade tasks

BjoernT:
The groups mons_hosts and cephrgwdummy_hosts are excluded
from the pre certain upgrade tasks as these nodes
are managed by ceph-ansible rather then OSA

Closes-Bug: Fleek-260


antonym:
recheck_all

PR #317, opened 2019-01-21
FLEEK-261 Remove ceph roles from older releases being tested

antonym:


antonym:

PR #321, opened 2019-01-25
Enable tempest for all, drop tempest ips

antonym:
MNAIO and AIO should specify their own tempest
IPs to use, so drop the hardcode and see what jobs
can pass with this change.

### DIFF
--- a/gating/check/run
+++ b/gating/check/run
@@ -114,7 +114,7 @@ function post_upgrade {
 }
 
 function run_tempest {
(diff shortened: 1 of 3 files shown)


antonym:

antonym:

antonym:
tempest still seems unstable on older releases so limiting to the incremental upgrades as they appear to be passing on a multi-node for N2P and P2Q.

PR #326, opened 2019-01-30
Bootstrap OSA if it hasn't been done before

antonym:
Ensures OSA bootstrap is ran to overwrite any RPC-O
specific bootstraps.  Writes a file to ensure it's
done at least once to flip things over so that we
always have OSA openstack-ansible available.

### DIFF
--- a/incremental/ubuntu16-upgrade-to-rocky.sh
+++ b/incremental/ubuntu16-upgrade-to-rocky.sh
@@ -30,6 +30,7 @@ echo "Starting Queens to Rocky Upgrade..."
 
 checkout_rpc_openstack
 configure_rpc_openstack
+ensure_osa_bootstrap
 prepare_rocky
 run_upgrade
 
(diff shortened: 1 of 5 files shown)


incremental/ubuntu16-upgrade-to-rocky.sh
```diff
@@ -28,6 +28,7 @@ export RPC_ANSIBLE_VERSION="2.5.5"
 
 echo "Starting Queens to Rocky Upgrade..."
 
+check_for_osa_bootstrap
```
rexredinger: Minor nit: Rename to ensure_osa_bootstrap, since it's not just checking but also bootstrapping if needed.

antonym: Good call, changes pushed.

antonym:

antonym:
recheck_all

antonym:

PR #327, opened 2019-01-31
FLEEK-268 Check for apt_pinned_packages in overrides

antonym:
apt_pinned_packages can break things during upgrades as
it overrides all pins in all roles.

This will fail if that variable is present prompting the user
to investigate and remove the setting if present.


antonym:

antonym:

PR #342, opened 2019-02-19
Clean out kilo submodules of ceph

antonym:
Kilo branches have ceph submodules built it,
so we purge them all after we check out so we
ensure osa is checked out without failure.


antonym:
recheck_container_ansible-lint_tox-test

PR #345, opened 2019-02-26
RI-508 - Run lxc-veth-check instead of restarting

antonym:
Attempts to get around containers loosing dhcp because
of lxc-dnsmasq transition without restarting containers


antonym:

PR #346, opened 2019-02-27
FLEEK-203 Ensure ca-certificates is latest

antonym:
During preflight, ensure ca-certificates package is
latest on hosts to avoid cert issues during leap.


antonym:

PR #350, opened 2019-03-19
Remove rpc-repo GPG keys

antonym:
Removes rpc-repo keys as they aren't used past Newton and prevents
apt update failures from expired keys.


antonym:
Seems that the cached container contains the rpco.list and keys so we'll prob need to regenerate it at least for ocata as it appears after the repo container teardown and rebuild.

PR #351, opened 2019-03-20
Drop out rpc-o specifics and use OSA for upgrades

antonym:
Removes RPC-O specific parts and relies on upstream for checking out each stable version of OSA to walk through the incremental upgrades.  Utilizes group_vars and env.d files from each upstream OSA release.  Ensures we're using the latest OSA stable branches for upgrades instead of relying on the outdated RPC-O builds.

Adds inspector for env.d files and alerts user if there are custom settings.  It purges any that match the release and leaves the ones that don't remaining for the operator to adjust.  The operator can then choose to skip the check if they are comfortable with those changes.

### DIFF
--- a/incremental/lib/functions.sh
+++ b/incremental/lib/functions.sh
@@ -117,21 +117,6 @@ function check_user_variables {
   fi
 }
 
-function checkout_rpc_openstack {
-  if [ ! -d "/opt/rpc-openstack" ]; then
-    git clone --branch ${RPC_BRANCH} --recursive https://github.com/rcbops/rpc-openstack /opt/rpc-openstack
-  else
-    pushd /opt/rpc-openstack
-      git clean -df
-      git reset --hard HEAD
-      git fetch --all
-      rm -rf openstack-ansible
-      rm -rf scripts/artifacts-building/
-      git checkout ${RPC_BRANCH}
-    popd
-  fi
-}
-
 function checkout_openstack_ansible {
   if [ ! -d "/opt/openstack-ansible" ]; then
     git clone --recursive https://github.com/openstack/openstack-ansible /opt/openstack-ansible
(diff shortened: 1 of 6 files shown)


antonym:
recheck_all

antonym:

antonym:

antonym:

antonym:

antonym:
recheck_all

antonym:

antonym:
@BjoernT see any issues with this?  this will enable us to use the latest stable from OSA upstream instead of the RPC-O versions which are probably falling out of date.

It'll also check for and stop the process if there are any differences to the env.d files that need to be mitigated before the upgrade.

incremental/lib/functions.sh
```diff
@@ -170,13 +155,11 @@ function ensure_osa_bootstrap {
...
 
-function configure_rpc_openstack {
-  rsync -av --delete /opt/rpc-openstack/etc/openstack_deploy/group_vars /etc/openstack_deploy/
-  rm -rf /opt/rpc-ansible
-  virtualenv /opt/rpc-ansible
-  install_ansible_source
-  pushd /opt/rpc-openstack/playbooks
-    /opt/rpc-ansible/bin/ansible-playbook -i 'localhost,' site-release.yml
+function configure_osa {
+  rm -rf /etc/openstack_deploy/group_vars
```
BjoernT: That's my biggest issue around group_vars and conf.d and overrides. I think we need to see if migrate these to a structure like https://github.com/rpc-environments/RPCO-OSA-Template/tree/master/global/configs/openstack to check if RPC default values are set for the upgrade

antonym: I'm fine with getting to that as an end goal, but for the most part, the group_vars in RPC-O by default contain things like:

Pike:
env.d/neutron.yml - For bare metal
group_vars/all/rpc-o.yml - horizon png overrides (don't need)
group_vars/all/mariadb.yml - Pin to galera repo (don't need)
group_vars/all/osa.yml - Various config overrides
group_vars/all/release.yml - rpc-release versions (don't need)

Queens:

No env.d specified
Same group_vars minus mariadb pin

Rocky:

Same as queens

Are there any other overrides we require?  Because we're ideally migrating to the newest versions, is it worth the hassle of getting those situated per increment?  For the most part, other than the env/neutron.yml and osa.yml (which could just use the OSA defaults) it seems like its relatively safe to remove those.

I'm not sure we could just drop in the RPCO-OSA-Template based on stein until we get the environment to a certain point given the variations in configs on each version but we could add logic to check for /opt/rpc-config to be populated before incrementing to a certain version and halt if it's not there for transitions to rocky/stein.  Then if rpc-config is present, we can ask if the operator wants to symlink openstack_deploy to rpc-config instead and then we'd use that for the upgrade.

antonym: @BjoernT pushed up a change... decided to just remove those files directly listed above that were part of the RPC-O releases and still do the env.d check.  That should get the rpc-o specific overrides out of the way that we don't want while leaving everything else as is.  If there are env.d changes required, it'll still halt to ask the user to check and override if needed.

Then once they are on rocky/stein, we cut over to the new config deployment template.

antonym:
Was merged in https://gitlab.com/rpco-eng/rpc-upgrades/merge_requests/8, closing.

PR #358, opened 2019-04-15
Preload apt packages (download, not install)

antonym:
Playbook that ensures apt is in a healthy state, updates cache
and then downloads apt packages for hosts and computes.

Can be ran ahead of a maintenance to prevent download bottlenecks
during the maintenance when time is more valuable.

Should work for both leapfrogs and incrementals, see readmes for each type of upgrade.


playbooks/preload-apt-newton.yml
```diff
@@ -0,0 +1,152 @@
...
+      - python-dev
+      - python3-lxc
+      - pxz
+  tasks:
+    - name: Download apt packages but don't install
+      shell: "apt-get -d -y install {{ item }}"
+      with_items:
+        - "{{ lxc_hosts_distro_packages | join(' ') }}"
+
+- name: Preload compute_host apt packages
```
BjoernT: This is a problem as the new cloud archives for newton are not laid down yet so the packaged you would download are actually not the ones you install

antonym: It's mainly just doing a download only without the install (apt-get -d), so we could just add the repo, and then retrieve the files directly so they are sitting in the cache ready to go.  This could also just be done on the deployment host where we download all of the files into the cache and sync it around ahead of time.  In the larger host environments, it seems like this would be a useful operation to reduce time so that it's not chewing up time during a maintenance window.  That or set up an apt cache proxy.

Mainly just was kicking around ideas on how to avoid us slamming the mirror during the deployment... lots of things should be able to be done so that we can at least reduce the time of that step.  I think if we could just get something working for just the computes in large environments that would probably help a bit.

BjoernT: Yepp I got that but the problem at this stage you are either missing the UCA or have older UCA repos deployed. This means you're not downloading the right package setup-hosts would do

antonym: @BjoernT Made some tweaks, if you specify the target release, i.e newton and then it'll install the appropriate repo, download the files using apt, and the removes the repo added.  The files seem to remain in place though so if the repo is re-added they'll be there ready to use.

cd /opt/rpc-upgrades/playbooks
openstack-ansible preload-apt-packages.yml -e target_release=newton

It'll also let us define the vars based on release/os/version so that we can preload whatever is typically loaded during the upgrade.

Obviously, this doesn't help us on containers package gets but there are much fewer of those since they live on infras.  This is mainly geared toward large number of computes but it should help reduce the time we're stuck in the os-nova-install playbook.  I have some more tuning to do with newer versions, but it should be usable to test on Newton. (tested with Ansible 1.9.4, had to let it assume the filename for apt_repository, newer versions let you specify filename).

antonym:
recheck_all

antonym:
recheck_all

PR #360, opened 2019-05-06
Rebuild LXC containers only once

BjoernT:
The LXC containers are only rebuild once which
allows for restart of the upgrade.sh script


antonym:
recheck_container_bashate_tox-test

PR #361, opened 2019-05-13
Use latest leapfrog changes:

antonym:
Brings in:
Skip neutron rebalance if ha-tool is disabled
https://review.opendev.org/#/c/652205/


antonym:
Commit looks good on L2N test, others failed from other gating issues, so merging in.

PR #365, opened 2019-05-28
Ensure openstack-release.upgrade is updated

antonym:
Ensures that openstack-release.upgrade reflects the completed
upgrade to avoid restarting from the original upgrade point.


antonym:

PR #367, opened 2019-05-31
Inspect files in env.d for deviations

antonym:
Pulls in check code from OSA stein to inspect env.d
files.  If there are modifications, user is prompted to check
and if they choose to ignore they have to:

export SKIP_CUSTOM_ENVD_CHECK=true


antonym:

antonym:
Moving to https://github.com/rcbops/rpc-upgrades/pull/351 commit since it uses OSA only.

PR #371, opened 2019-06-11
Adds additional reports for checking status

antonym:
Checks list of endpoints to ensure lb addreses match  appropriate endpoints and lists in file.

Adds osa_health role for gathering data from all hosts and generates a TLDR report that can be easily parsed. The role currently generates a list of updates that need to be done on all containers/hosts and does a check for zombie processes.

Renames status file to openstack-status.


antonym:
Merged in https://gitlab.com/rpco-eng/rpc-upgrades/merge_requests/9

PR #372, opened 2019-06-12
Add Rocky prep to incremental

antonym:
Clears out apt proxies and corrects working directory for
existing queens prep.


antonym:
recheck_all

antonym:
@BjoernT not sure if this is what you ran into on your Q2R upgrade, but found this fixes an issue I was having testing Q2R.

PR #373, opened 2019-06-14
Adds initial support for upgrades to stein/train

antonym:
Adds initial support for upgrading to stein/train and treats rocky
as the upgrade point from Ubuntu 16.04 to 18.04 to test
with.  Anything being upgraded from 18.04 will make the
assumption that the rpc-config directory is required.

Rocky upgrades and up expect the rpc-config directory
to be present and setup (checks if /opt/rpc-config exists). If the /opt/rpc-config is not present, the upgrade will halt on stein/train upgrades.

### DIFF
--- a/incremental/ubuntu18-upgrade-to-stein.sh
+++ b/incremental/ubuntu18-upgrade-to-stein.sh
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
 
-# Copyright 2018, Rackspace US, Inc.
+# Copyright 2019, Rackspace US, Inc.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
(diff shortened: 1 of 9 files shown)


incremental/ubuntu18-upgrade-to-stein.sh
```diff
@@ -0,0 +1,35 @@
...
+
+source lib/functions.sh
+source lib/vars.sh
+
+require_ubuntu_version 18
+
+export OSA_SHA="stable/stein"
+export SKIP_INSTALL=${SKIP_INSTALL:-'no'}
+export RPC_PRODUCT_RELEASE="stein"
+export RPC_ANSIBLE_VERSION="2.7.8"
```
BjoernT: OSA bumped to 2.7.9

antonym: Did they?  I pulled the current version # from here: https://github.com/openstack/openstack-ansible-tests/blob/stable/stein/test-ansible-deps.txt#L11

BjoernT: Yes https://github.com/openstack/openstack-ansible/blob/stable/stein/scripts/bootstrap-ansible.sh#L25

So they test something different they install, fantastic

antonym: Guess it's a mismatch, https://github.com/openstack/openstack-ansible/blob/stable/stein/scripts/bootstrap-ansible.sh#L25 heh, k will update and merge.

antonym:
Updated and merging... the gating systems all busted because of JIRA so I removed requirements on all jobs.

PR #381, opened 2020-01-24
Convert rsync tasks to one-time sync

BjoernT:
The rsync tasks of the group vars is converted into a one-time
task to allow edits during upgrades without destroying these
upon restart


shannonmitchell:
recheck_all

shannonmitchell:

PR #382, opened 2020-03-31
Update RPC_BRANCH for rocky incremental upgrades to a tagged version

BjoernT:


gating/check/pre
```diff
@@ -38,7 +38,8 @@ ${GETPIP_CMD} https://bootstrap.pypa.io/3.3/get-pip.py > ${GETPIP_FILE}\
     || ${GETPIP_CMD} https://raw.githubusercontent.com/pypa/get-pip/master/3.3/get-pip.py > ${GETPIP_FILE}
 
 # Install specific verison of pip/setuptools/wheel.
-sudo ${GETPIP_PYTHON_EXEC_PATH} ${GETPIP_FILE} 'pip==9.0.3' 'setuptools==39.0.1' 'wheel==0.30.0'
+# virtualenv >= 20 seems to trigger https://github.com/pypa/virtualenv/issues/1638
+sudo ${GETPIP_PYTHON_EXEC_PATH} ${GETPIP_FILE} 'pip==9.0.3' 'setuptools==39.0.1' 'wheel==0.30.0' 'virtualenv<20'
```
BjoernT, line 42: @shannonmitchell Do you know off hand if any tool is installing a custom site.py leading to the issue per https://github.com/pypa/virtualenv/issues/1638 ? I worked around with limiting virtualenv to <20

shannonmitchell: I do not.  I haven't had enough time to dive into the specifics on any of these other than looking into the random brokenness on the jenkins infrastructure side of things.

BjoernT: I think I tracked it down to setuptools and a --no-setuptools my be ok too but the limits on virtualenv < 20 seems reasonable to me for now, at least for all old releases until train. Python 3 requires >=20 in the end. The other issues I saw are all related to nodepool as that one is down so instances are not created

PR #396, opened 2021-05-14
Update variable-migrations.yml

brewc:
fix conditional error


BjoernT:
Same as #397 this should not be an issue per https://docs.ansible.com/ansible/latest/user_guide/playbooks_conditionals.html as is success replaced |success

BjoernT:
Closing with comment above

